PR #107, opened 2021-12-26
Proposal for new approach with nanoseconds integers

pwnorbitals:
Did some light refactoring and benchmarked (will be deleted before merging). What are your thoughts about this approach ?

Note : this currently relies on bigint_helper_methods (https://github.com/rust-lang/rust/issues/85532) so compiles on Nightly only

### DIFF
--- a/src/duration.rs
+++ b/src/duration.rs
@@ -14,11 +14,17 @@ use std::fmt;
 use std::ops::{Add, AddAssign, Div, Mul, Neg, Sub, SubAssign};
 use std::str::FromStr;
 
-const DAYS_PER_CENTURY_U: u128 = 36_525;
-const SECONDS_PER_MINUTE_U: u128 = 60;
-const SECONDS_PER_HOUR_U: u128 = 3_600;
-const SECONDS_PER_DAY_U: u128 = 86_400;
-const ONE: u128 = 1_u128;
+
+
+const SECONDS_PER_MINUTE_U: u64 = 60;
+const MINUTES_PER_HOUR_U: u64 = 60;
+const HOURS_PER_DAY_U: u64 = 24;
+const SECONDS_PER_HOUR_U: u64 = SECONDS_PER_MINUTE_U * MINUTES_PER_HOUR_U;
+const SECONDS_PER_DAY_U: u64 = SECONDS_PER_HOUR_U * HOURS_PER_DAY_U;
+const DAYS_PER_CENTURY_U: u64 = 36_525;
+const NS_PER_DAY_U: u64 = 1e9 as u64 * SECONDS_PER_DAY_U;
+const NS_PER_CENTURY_U: u64 = DAYS_PER_CENTURY_U * NS_PER_DAY_U;
+const ONE: u64 = 1_u64;
 
 /// Defines generally usable durations for high precision math with Epoch (all data is stored in seconds)
 #[derive(Clone, Copy, Debug, PartialEq, PartialOrd)]
--- a/src/duration.rs
+++ b/src/duration.rs
@@ -14,15 +14,26 @@ use std::str::FromStr;
 
 
 
-const SECONDS_PER_MINUTE_U: u64 = 60;
-const MINUTES_PER_HOUR_U: u64 = 60;
-const HOURS_PER_DAY_U: u64 = 24;
-const SECONDS_PER_HOUR_U: u64 = SECONDS_PER_MINUTE_U * MINUTES_PER_HOUR_U;
-const SECONDS_PER_DAY_U: u64 = SECONDS_PER_HOUR_U * HOURS_PER_DAY_U;
-const DAYS_PER_CENTURY_U: u64 = 36_525;
-const NS_PER_DAY_U: u64 = 1e9 as u64 * SECONDS_PER_DAY_U;
-const NS_PER_CENTURY_U: u64 = DAYS_PER_CENTURY_U * NS_PER_DAY_U;
+const SECONDS_PER_MINUTE_U: u16 = 60;
+const MINUTES_PER_HOUR_U: u16 = 60;
+const HOURS_PER_DAY_U: u16 = 24;
+const SECONDS_PER_HOUR_U: u16 = SECONDS_PER_MINUTE_U * MINUTES_PER_HOUR_U;
+const SECONDS_PER_DAY_U: u32 = SECONDS_PER_HOUR_U as u32 * HOURS_PER_DAY_U as u32;
+const DAYS_PER_CENTURY_U: u16 = 36_525;
+const NS_PER_DAY_U: u64 = 10u64.pow(9) * SECONDS_PER_DAY_U as u64;
+const NS_PER_CENTURY_U: u64 = DAYS_PER_CENTURY_U as u64 * NS_PER_DAY_U as u64;
 const ONE: u64 = 1_u64;
+
+const SECONDS_PER_MINUTE_F: f64 = 60.0;
+const MINUTES_PER_HOUR_F: f64 = 60.0;
+const HOURS_PER_DAY_F: f64 = 24.0;
+const SECONDS_PER_HOUR_F: f64 = SECONDS_PER_MINUTE_F * MINUTES_PER_HOUR_F;
+const SECONDS_PER_DAY_F: f64 = SECONDS_PER_HOUR_F * HOURS_PER_DAY_F;
+const DAYS_PER_CENTURY_F: f64 = 36_525.0;
+const NS_PER_DAY_F: f64 = 1e9 * SECONDS_PER_DAY_F;
+const NS_PER_CENTURY_F: f64 = DAYS_PER_CENTURY_F * NS_PER_DAY_F;
+const ONE_F: f64 = 1.0;
+
 
 /// Defines generally usable durations for high precision math with Epoch
 #[derive(Clone, Copy, Debug, PartialEq, PartialOrd)]
@@ -51,15 +62,15 @@ macro_rules! impl_ops_f {
             fn mul(self, q: $type) -> Duration {
                 match self {
                     TimeUnit::Century => {
-                        Duration::from_days_f(q as f64 * DAYS_PER_CENTURY_U as f64)
+                        Duration::from_days_f(f64::from(q) * DAYS_PER_CENTURY_F)
                     }
-                    TimeUnit::Day => Duration::from_days_f(q as f64),
-                    TimeUnit::Hour => Duration::from_hours_f(q as f64),
-                    TimeUnit::Minute => Duration::from_minutes_f(q as f64),
-                    TimeUnit::Second => Duration::from_seconds_f(q as f64),
-                    TimeUnit::Millisecond => Duration::from_milliseconds_f(q as f64),
-                    TimeUnit::Microsecond => Duration::from_microseconds_f(q as f64),
-                    TimeUnit::Nanosecond => Duration::from_nanoseconds_f(q as f64),
+                    TimeUnit::Day => Duration::from_days_f(f64::from(q)),
+                    TimeUnit::Hour => Duration::from_hours_f(f64::from(q)),
+                    TimeUnit::Minute => Duration::from_minutes_f(f64::from(q)),
+                    TimeUnit::Second => Duration::from_seconds_f(f64::from(q)),
+                    TimeUnit::Millisecond => Duration::from_milliseconds_f(f64::from(q)),
+                    TimeUnit::Microsecond => Duration::from_microseconds_f(f64::from(q)),
+                    TimeUnit::Nanosecond => Duration::from_nanoseconds_f(f64::from(q)),
                 }
             }
         }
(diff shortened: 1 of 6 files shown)


ChristopherRabotin:
Thanks for the pr!

If I read the benchmarks correctly, using nanos and centuries is about 3 times faster than two floats for the `TBD seconds and JDE ET` test, 8 times _slower_ for `Duration to f64 seconds time`, about 40% faster for `Duration add and assert day hour`, marginally _slower_ for `Duration add and assert minute second`, and like 20% slower for `Duration add and assert subsecons`.

Also, for the record, the conversion was about whether an FPU was available on embedded platforms, right? And that, typically, an FPU is slower than a APU. Was there another reason?

pwnorbitals:
Just updated the PR. Indeed, going to f64 is slower but all other operations in the end should be faster. For now, I use a 128-bit type as intermediary for the addition operation, which should not be needed when optimizing further as we could switch to a `carrying_add` with a bit of additional logic, which should gain some performance. Same for substraction.

Apart for performance, advantages are compatibility with platforms without FPU and guaranteed nanosecond-precision anywhere in the possible Duration range (implementations using floating-points have variable time precision across the range, with very low precision at the extremes).

pwnorbitals:
Cleaned-up the branch a bit, almost all tests are passing now. Still need some work on the printing side, and it looks like there might be some nasty rounding errors I need to track down, but I'm positive this gets finalized soon™

pwnorbitals:
r ? @ChristopherRabotin

src/duration.rs
```diff
@@ -658,8 +869,8 @@ fn time_unit() {
     let sum: Duration = seven_hours + six_minutes + five_seconds;
     assert!((sum.in_seconds() - 25565.0).abs() < EPSILON);
 
-    let neg_sum = -sum;
-    assert!((neg_sum.in_seconds() + 25565.0).abs() < EPSILON);
+    let neg_sum = dbg!(-dbg!(sum));
```
ChristopherRabotin, line 872: Remove `dbg!`

src/duration.rs
```diff
@@ -681,81 +892,73 @@ fn time_unit() {
...
     assert!((sum.in_unit_f64(TimeUnit::Minute) + 35.0).abs() < EPSILON);
 }
 
 #[test]
 fn duration_print() {
     // Check printing adds precision
     assert_eq!(
-        format!("{}", TimeUnit::Day * 10.0 + TimeUnit::Hour * 5),
-        "10 days 5 h 0 min 0 s"
+        format!("{}", TimeUnit::Day * 10.0 + TimeUnit::Hour * 5).trim(),
```
ChristopherRabotin, line 904: Why `trim()`?

src/duration.rs
```diff
@@ -681,81 +892,73 @@ fn time_unit() {
...
+
     assert_eq!(
         format!(
             "{}",
-            TimeUnit::Hour * -5 + TimeUnit::Millisecond * -256 + TimeUnit::Nanosecond * -3.5
-        ),
-        "-5 h 0 min 0 s 256 ms 3.5 ns"
+            TimeUnit::Hour * -5 + TimeUnit::Millisecond * -256 + dbg!(TimeUnit::Nanosecond * -3)
+        ).trim(),
+        "-5 h 0 min 0 s 256 ms 0 us 3 ns"
```
ChristopherRabotin, line 940: Nice! This is better I think

src/duration.rs
```diff
@@ -681,81 +892,73 @@ fn time_unit() {
...
             (TimeUnit::Hour * -5 + TimeUnit::Millisecond * -256)
                 - (TimeUnit::Hour * -5 + TimeUnit::Millisecond * -256 + TimeUnit::Nanosecond * 2)
-        ),
+        ).trim(),
         "-2 ns"
     );
 
     assert_eq!(
-        format!("{}", Duration::from_nanoseconds(TwoFloat::from(2))),
+        format!("{}", Duration::from_nanoseconds_f(2.0)).trim(),
```
ChristopherRabotin, line 953: I wonder if the `from_nanoseconds` shouldn't be generic over some num_trait type to avoid the `_f` and `_u`. Thoughts?

src/duration.rs
```diff
@@ -681,81 +892,73 @@ fn time_unit() {
...
     );
 
     // Check that we support nanoseconds pas GPS time
-    let now = TimeUnit::Nanosecond * 1286495254000000123_u128;
+    let now = TimeUnit::Nanosecond * 1286495254000000203_u128;
     assert_eq!(
-        format!("{}", now),
-        "14889 days 23 h 47 min 34 s 0 ms 203.125 ns"
+        format!("{}", now).trim(),
+        "40 years 289 days 23 h 47 min 34 s 0 ms 0 us 203 ns"
```
ChristopherRabotin, line 961: Hmm, why is this test different? Actually, shouldn't the original test ending with 123ns actually display 203 nanoseconds as you do? Seems like a bug in the previous version

src/duration.rs
```diff
@@ -58,85 +68,344 @@ macro_rules! impl_ops_for_type {
...
+    pub fn new(ns : u64, centuries : i64) -> Self {
+        let mut out = Duration { ns, centuries };
+        out.normalize();
+        out
     }
-    pub fn from_hours(hours: TwoFloat) -> Self {
-        Self {
-            0: hours * TwoFloat::from(SECONDS_PER_HOUR_U),
+
+    pub fn total_ns(self) -> i128 {
```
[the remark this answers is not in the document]
ChristopherRabotin: Maybe the switch to J2000 should be done in another PR to merge this one in faster.

src/duration.rs
```diff
@@ -765,8 +968,8 @@ fn duration_print() {
         + 0.milliseconds()
         + 123.nanoseconds();
     assert_eq!(
-        format!("{}", arbitrary),
-        "14889 days 23 h 47 min 34 s 0 ms 123 ns"
+        format!("{}", arbitrary).trim(),
+        "40 years 289 days 23 h 47 min 34 s 0 ms 0 us 123 ns"
```
ChristopherRabotin, line 972: Are you sure this test is correct? Can we add more `years` tests because I'm just worried we'll get the conversion wrong.

ChristopherRabotin (changes requested):
Very good job! I think we (you) should change centuries to a smaller representation because we don't need lots of bits there. Also, avoid the overflows with the `::from()` call instead of coercing the conversion where possible.

src/duration.rs
pwnorbitals: Because my Display implementation adds a space at the end of the string and I thought I'd rather trim() the tests than add complexity to the impl just for a single space

src/duration.rs
pwnorbitals: I assumed 1 year = 365 days, as this is only used for Display. It's easy to rollback the `years` part of the decomposition if you're not comfortable with it, I just thought I'd roll with it as it was basically free in the new Display impl. We could indeed add more `years` tests though

src/duration.rs
```diff
@@ -1,28 +1,36 @@
...
+const NS_PER_DAY_U: u64 = 1e9 as u64 * SECONDS_PER_DAY_U;
+const NS_PER_CENTURY_U: u64 = DAYS_PER_CENTURY_U * NS_PER_DAY_U;
+const ONE: u64 = 1_u64;
+
+/// Defines generally usable durations for high precision math with Epoch
 #[derive(Clone, Copy, Debug, PartialEq, PartialOrd)]
-pub struct Duration(TwoFloat);
+pub struct Duration {
+    ns : u64, // 1 century is about 3.1e18 ns, and max value of u64 is about 1e19.26
+    centuries : i64 // +- 9.22e18 centuries is the possible range for a Duration
```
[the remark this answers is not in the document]
pwnorbitals: Agreed, will edit

src/duration.rs
ChristopherRabotin: Sounds good, we can make this a new issue.

src/duration.rs
ChristopherRabotin: I think that the IETF or someone defines a year as 365.25 days (since one century is 36525 days). But the problem is that it's an approximate definition, and that's why lots of implementations don't allow for year based computations: it's confusing.

pwnorbitals: Understood, it makes sense, I will remove this part of the decomposition then

ChristopherRabotin:
Okay, I did a few changes to this branch. It doesn't work yet, but there's progress: I unified the `ops` macro so that we don't have to specify what is the type needed. Tests fails pretty badly but the fact it compiles is a start!

```
test duration::deser_test ... ok
test epoch::deser_test ... ok
test epoch::datetime_invalid_dates ... ok
test epoch::leap_year ... ok
test epoch::quorem_nominal_test ... ok
test duration::duration_print ... FAILED
test duration::time_unit ... FAILED
test epoch::quorem_nil_den_test - should panic ... ok
test epoch::gpst ... FAILED
test epoch::julian_epoch ... FAILED
test epoch::regression_test_gh_85 ... FAILED
test epoch::test_range ... ok
test epoch::spice_et_tdb ... FAILED
test error_unittest ... ok
test epoch::utc_tai ... FAILED
test epoch::ops ... FAILED
test epoch::utc_epochs ... FAILED
test timeseries::test_timeseries ... FAILED
test sim::clock_noise_up ... ok
test sim::clock_sample ... ok
test epoch::test_from_str ... FAILED

```

ChristopherRabotin:
Just posting some temporary benchmarks from this PR after I started working on it last week. 

**Summary:**
+ Converting a duration to `f64` is almost **three times slower**: 10.29 ns vs 3.40 ns
+ Every other operation, including converting an epoch to its JDE seconds or TDB seconds (used for ephemeris querying) is up to **twice as fast**

&nbsp; | fixed nanosecond precision | TwoFloat 
--|---|--
TBD seconds and JDE ET | **430.39 ns** | 1.0321 us
Duration to f64 seconds | 10.480 ns | **3.4000 ns**
Duration add and assert day hour | **24.747 ns** | 68.897 ns
Duration add and assert minute second | **22.536 ns** | 32.867 ns
Duration add and assert subseconds | **34.316 ns** | 32.319 ns

### With nanosecond precision
```
     Running unittests (target/release/deps/bench_epoch-683b87af3f4470f4)
TBD seconds and JDE ET  time:   [424.62 ns 430.39 ns 437.20 ns]                                   
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) high mild
  2 (2.00%) high severe

Duration to f64 seconds time:   [10.404 ns 10.480 ns 10.569 ns]                                     
Found 8 outliers among 100 measurements (8.00%)
  6 (6.00%) high mild
  2 (2.00%) high severe

Duration add and assert day hour                                                                             
                        time:   [24.605 ns 24.747 ns 24.917 ns]
Found 9 outliers among 100 measurements (9.00%)
  4 (4.00%) low mild
  3 (3.00%) high mild
  2 (2.00%) high severe

Duration add and assert minute second                                                                             
                        time:   [22.437 ns 22.536 ns 22.636 ns]
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild
  1 (1.00%) high severe

Duration add and assert subsecons                                                                             
                        time:   [34.173 ns 34.316 ns 34.464 ns]
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) low mild
  2 (2.00%) high mild
  2 (2.00%) high severe
```

### With TwoFloat
Note: this was executed _after_ the nanosecond precision so the "performance" increase/decrease metrics are compared to the nanosecond branch

```
test result: ok. 0 passed; 0 failed; 21 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests (target/release/deps/bench_epoch-826101cd40d0f022)
TBD seconds and JDE ET  time:   [1.0265 us 1.0321 us 1.0376 us]                                    
                        change: [+138.56% +140.95% +143.07%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild

Duration to f64 seconds time:   [3.3858 ns 3.4000 ns 3.4155 ns]                                     
                        change: [-68.371% -67.860% -67.461%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high mild

Duration add and assert day hour                                                                            
                        time:   [68.197 ns 68.897 ns 69.607 ns]
                        change: [+178.35% +181.05% +183.91%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) low mild
  2 (2.00%) high mild

Duration add and assert minute second                                                                             
                        time:   [32.746 ns 32.867 ns 32.982 ns]
                        change: [+44.195% +45.015% +45.828%] (p = 0.00 < 0.05)
                        Performance has regressed.

Duration add and assert subsecons                                                                             
                        time:   [32.172 ns 32.319 ns 32.462 ns]
                        change: [-7.0698% -6.4217% -5.7948%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
```

ChristopherRabotin:
Crap... just implement a bunch of code fixes and now the performance is quite bad

ChristopherRabotin:
Ha! Switching the creation of a `Duration` from initializing an i64 instead of an i128 where possible greatly decreased execution time. It introduces a few bugs I need to iron out, but this is very promising.

```
test result: ok. 0 passed; 0 failed; 24 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests (target/release/deps/bench_epoch-1b18a789a5c33cd2)
TBD seconds and JDE ET  time:   [328.74 ns 331.58 ns 334.83 ns]                                   
                        change: [-60.090% -59.719% -59.288%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  3 (3.00%) high mild
  4 (4.00%) high severe

Duration to f64 seconds time:   [5.9576 ns 6.0285 ns 6.1116 ns]                                     
                        change: [-68.638% -68.360% -68.052%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe

Duration add and assert day hour                                                                             
                        time:   [14.576 ns 14.788 ns 15.032 ns]
                        change: [-67.917% -67.200% -66.523%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  5 (5.00%) high mild
  5 (5.00%) high severe

Duration add and assert minute second                                                                             
                        time:   [14.172 ns 14.249 ns 14.333 ns]
                        change: [-66.959% -66.591% -66.261%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild

Duration add and assert subsecons                                                                             
                        time:   [13.948 ns 14.002 ns 14.060 ns]
                        change: [-66.207% -65.628% -65.081%] (p = 0.00 < 0.05)
                        Performance has improved.

```

ChristopherRabotin:
Final benchmarks

&nbsp; | fixed nanosecond precision | TwoFloat 
--|---|--
TBD seconds and JDE ET | **384.88 ns** | 1.0321 us
Duration to f64 seconds | 6.1890 ns | **3.4000 ns**
Duration add and assert day hour | **14.809 ns** | 68.897 ns
Duration add and assert minute second | **15.781 ns** | 32.867 ns
Duration add and assert subseconds | **15.804 ns** | 32.319 ns

ChristopherRabotin (approved):

PR #110, opened 2022-02-08
New methods and a fix for the GPS epoch

cjordan:
Hi!  Thanks very much for `hifitime`, it's been a lifesaver to deal with JDs, MJDs and leap seconds.  I've been using it for about 1.5 years now.

I've got a lot going on in this PR so if you'd prefer for me to break it up or you have any questions, please let me know.  I also don't 100% trust myself with getting everything correct so I'd appreciate some experienced eyes over the changes.

The first commit is simple enough - it just specifies patch versions for dependencies seeing as that's generally a good thing to do.  In the commit message I've linked the relevant Rust users forum post that set me off tidying up this kinda stuff all over the place.

The second commit adds a bunch of functions that seemed to simply be missing from the library previously (`from_utc_{seconds,days}`, `from_{mjd,jde}_utc` and `from_gpst_{seconds,days}`). I'm reasonably happy with them but I'm not certain that everything is correct.

The third commit is the most significant and if this PR is merged, a new release of `hifitime` should probably be `3.0.0` because it's a breaking change.  Basically, it puts the GPS epoch at 1980/1/5 rather than just 19 seconds after TAI.  This helps me a lot because I frequently use GPS time for work and I had to use a hack to get around this in the past.

Your time and help is much appreciated, thanks!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -160,6 +160,14 @@ pub const SECONDS_PER_YEAR: f64 = 31_557_600.0;
 pub const SECONDS_PER_TROPICAL_YEAR: f64 = 31_556_925.974_7;
 /// `SECONDS_PER_SIDERAL_YEAR` corresponds to the number of seconds per sideral year from [NIST](https://www.nist.gov/pml/special-publication-811/nist-guide-si-appendix-b-conversion-factors/nist-guide-si-appendix-b9#TIME).
 pub const SECONDS_PER_SIDERAL_YEAR: f64 = 31_558_150.0;
+/// `SECONDS_GPS_TAI_OFFSET` is the number of seconds from the TAI epoch to the
+/// GPS epoch (UTC midnight of January 6th 1980; cf.
+/// https://gssc.esa.int/navipedia/index.php/Time_References_in_GNSS#GPS_Time_.28GPST.29)
+pub const SECONDS_GPS_TAI_OFFSET: f64 = 80.0 * SECONDS_PER_YEAR + 4.0 * SECONDS_PER_DAY + 19.0;
+/// `DAYS_GPS_TAI_OFFSET` is the number of days from the TAI epoch to the GPS
+/// epoch (UTC midnight of January 6th 1980; cf.
+/// https://gssc.esa.int/navipedia/index.php/Time_References_in_GNSS#GPS_Time_.28GPST.29)
+pub const DAYS_GPS_TAI_OFFSET: f64 = SECONDS_GPS_TAI_OFFSET / SECONDS_PER_DAY;
 
 mod sim;
 pub use sim::ClockNoise;
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,16 +12,16 @@ readme = "README.md"
 license = "Apache-2.0"
 
 [dependencies]
-rand = "0.8"
-rand_distr = "0.4"
-regex = "1.3"
-serde = "1.0"
-serde_derive = "1.0"
+rand = "0.8.0"
+rand_distr = "0.4.0"
+regex = "1.3.0"
+serde = "1.0.0"
+serde_derive = "1.0.0"
 twofloat = "0.4.1"
 
 [dev-dependencies]
-serde_derive = "1.0"
-criterion = "0.3"
+serde_derive = "1.0.0"
+criterion = "0.3.0"
 
 [[bench]]
 name = "bench_epoch"
--- a/src/epoch.rs
+++ b/src/epoch.rs
@@ -6,8 +6,8 @@ use self::regex::Regex;
 use self::serde::{de, Deserialize, Deserializer};
 use crate::duration::{Duration, TimeUnit};
 use crate::{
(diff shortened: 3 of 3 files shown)


src/lib.rs
```diff
@@ -160,6 +160,14 @@ pub const SECONDS_PER_YEAR: f64 = 31_557_600.0;
 pub const SECONDS_PER_TROPICAL_YEAR: f64 = 31_556_925.974_7;
 /// `SECONDS_PER_SIDERAL_YEAR` corresponds to the number of seconds per sideral year from [NIST](https://www.nist.gov/pml/special-publication-811/nist-guide-si-appendix-b-conversion-factors/nist-guide-si-appendix-b9#TIME).
 pub const SECONDS_PER_SIDERAL_YEAR: f64 = 31_558_150.0;
+/// `SECONDS_GPS_TAI_OFFSET` is the number of seconds from the TAI epoch to the
+/// GPS epoch (UTC midnight of January 5th 1980; cf.
+/// https://gssc.esa.int/navipedia/index.php/Time_References_in_GNSS#GPS_Time_.28GPST.29)
+pub const SECONDS_GPS_TAI_OFFSET: f64 = SECONDS_PER_YEAR * 80.0 + SECONDS_PER_DAY * 4.0;
```
ChristopherRabotin, line 166: This PR looks great, thanks! Just one question: I'm not sure I follow the math on the `SECONDS_GPS_TAI_OFFSET`. How does 4 days + 80 seconds corresponds to the January 5th 1980 UTC ? Note that I never deal with GPS epoch so I most likely am misunderstanding something crucial here and if I were to bet, I'd bet your math is correct.

From the reference:
> It starts at 0h UTC (midnight) of January 5th to 6th 1980 (6.d0). At that epoch, the difference TAI−UTC was 19 seconds.

Does that quote imply the offset should be 4 days and 19 seconds instead of 4 days and 80 seconds?

ChristopherRabotin:
Thanks for the PR, this is great work! It's always a pleasure to see other folks use this work instead of just me, especially seeing that you're using it on the MWAT radio telescope.

Good call for the version numbers in the Cargo.toml and the version bump.

In another PR, we're trying to change the `Duration` definition such that it has a fixed precision of one nanosecond for +/- 32,767 centuries around J1900: this would also enable using hifitime on embedded devices that don't have a floating point unit (FPU). That shouldn't lead to any breaking changes in `Epoch`, only some in `Duration`. So, two questions for you: 
1. Would you be fine if this PR changing the version of hifitime to `3.0.0-beta.1` and only bump to `3.0.0` once the other PR is fixed and merged?
2. In your work, do you use hifitime for better than nanosecond precision? I want to make sure that a flagship project like yours can still be supported.

src/lib.rs
cjordan: Thanks for this comment - it made me scrutinise things more closely. It is confusing, but the correct value is:

```rust
pub const SECONDS_GPS_TAI_OFFSET: f64 = 80.0 * SECONDS_PER_YEAR + 4.0 * SECONDS_PER_DAY + 19.0;
````
(same as before but +19s).  I think you missed the 80 being applied to seconds per *year*, so this is 80 years, 4 days and 19 seconds.  I was banging my head for a while thinking that this takes us to midnight of January 5th 1980, but I verified that this is right with some new tests:

```rust
    let gps_epoch = Epoch::from_tai_seconds(SECONDS_GPS_TAI_OFFSET);
    assert_eq!(
        gps_epoch.as_gregorian_str(TimeSystem::UTC),
        "1980-01-06T00:00:00 UTC"
    );
    assert_eq!(
        gps_epoch.as_gregorian_str(TimeSystem::TAI),
        "1980-01-06T00:00:19 TAI"
    );
```

Hopefully this is foolproof!

cjordan:
Thanks for taking the time to look at this, and I'm glad I could vindicate your work!

The nanosecond precision stuff sounds super cool, good luck with it.  I'd be scared to work on it because I find all this time stuff so confusing.

To answer your questions:

1. Yep, `3.0.0-beta.1` is absolutely fine with me.  I'm frankly just grateful to get a new release.
2. While precision is important to us, we're nowhere near the nanosecond level.  So your work in parallel doesn't affect us, that's all good.

I've force pushed a new commit for the GPS epoch stuff - it fixes a few little things like erroneously saying that the GPS epoch was January 5 instead of 6, and hopefully makes the tests clearer.  I've also verified stuff with astropy:

```rust
let gps_epoch = Epoch::from_tai_seconds(SECONDS_GPS_TAI_OFFSET);
dbg!(gps_epoch.as_jde_utc_days());
```
produces
```
[src/epoch.rs:1446] gps_epoch.as_jde_utc_days() = 2444244.5
```

and astropy says:
```python
from astropy.time import Time
from datetime import datetime
t = Time(datetime(1980, 1, 6), scale="utc")
t.gps >>> 0.0 
t.jd >>> 2444244.5
```

ChristopherRabotin:
Thanks for the astropy validation. The crate is up on crates.io: https://crates.io/crates/hifitime/3.0.0-beta.1 !

cjordan:
Thanks very much. All the best!

PR #120, opened 2022-05-02
Add UNIX timestamp

### LINKED ISSUE #22: Add Unix as a TimeSystem

mkolopanis:
Thanks for all the hard work you've done on this on this project. I've been using it for astronomy tools in my lab for about a year now.

I've added conversion function to/from the unix epoch timestamp (01 Jan 1970 00:00:00 UTC).
If you'd like I could add unix_tai convenience functions too which do not remove the leap seconds.

closes #22

### DIFF
--- a/src/epoch.rs
+++ b/src/epoch.rs
@@ -1,7 +1,7 @@
 use crate::duration::{Duration, Unit};
 use crate::{
     Errors, TimeSystem, DAYS_GPS_TAI_OFFSET, ET_EPOCH_S, J1900_OFFSET, J2000_OFFSET, MJD_OFFSET,
-    SECONDS_GPS_TAI_OFFSET, SECONDS_GPS_TAI_OFFSET_I64, SECONDS_PER_DAY, UNIX_OFFSET_UTC_SECONDS,
+    SECONDS_GPS_TAI_OFFSET, SECONDS_GPS_TAI_OFFSET_I64, SECONDS_PER_DAY,
 };
 use core::fmt;
 use core::ops::{Add, AddAssign, Sub, SubAssign};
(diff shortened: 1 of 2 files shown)


src/epoch.rs
```diff
@@ -1540,6 +1596,65 @@ fn gpst() {
...
 
+#[test]
+fn unix() {
+    use core::f64::EPSILON;
+    let now = Epoch::from_gregorian_utc_hms(2022, 5, 2, 10, 39, 15);
+    assert!(
+        now.as_tai_seconds() > now.as_utc_seconds(),
+        "TAI is not ahead of UTC."
+    );
+    assert!((now.as_tai_seconds() - now.as_utc_seconds() - 37.0).abs() < EPSILON);
```
ChristopherRabotin, line 1607: I think this test and the one above are from the TAI/UTC test and can be safely removed from this function.

src/epoch.rs
```diff
@@ -1540,6 +1596,65 @@ fn gpst() {
...
+    assert!(
+        now.as_tai_seconds() > now.as_unix_seconds(),
+        "TAI is not ahead of UNIX."
+    );
+    assert!((now.as_unix_seconds() - 1651487955.0f64).abs() < EPSILON);
```
ChristopherRabotin, line 1612: Minor: I typically like the underscore to specify the type, it makes it easier to read.
```suggestion
    assert!((now.as_unix_seconds() - 1651487955.0_f64).abs() < EPSILON);
```

src/epoch.rs
```diff
@@ -675,6 +704,33 @@ impl Epoch {
...
+        self.as_unix_duration().in_unit(unit)
+    }
+
+    /// Returns the number seconds since the UNIX epoch defined 01 Jan 1970 midnight UTC.
+    pub fn as_unix_seconds(self) -> f64 {
+        self.as_unix(Unit::Second)
+    }
+
+    /// Returns the number miliseconds since the UNIX epoch defined 01 Jan 1970 midnight UTC.
+    pub fn as_unix_miliseconds(self) -> f64 {
```
ChristopherRabotin, line 725: Milliseconds has two `l`s.
```suggestion
    pub fn as_unix_milliseconds(self) -> f64 {
```

src/epoch.rs
```diff
@@ -675,6 +704,33 @@ impl Epoch {
         self.as_gpst_duration().in_unit(Unit::Day)
     }
 
+    ///Returns the Duration since the UNIX epoch UTC midnight 01 Jan 1970.
+    pub fn as_unix_duration(self) -> Duration {
+        let cnt = self.get_num_leap_seconds();
+        // TAI = UNIX + leap_seconds + UNIX_OFFSET_UTC_SECONDS <=> UNIX = TAI - leap_seconds - UNIX_OFFSET_UTC_SECONDS
+        self.0 + i64::from(-cnt) * Unit::Second - UNIX_OFFSET_UTC_SECONDS * Unit::Second
```
ChristopherRabotin, line 711: Would this be easier if it were defined as an offset from UTC. From what I saw, the UNIX epoch is `January 1st, 1970 at 00:00:00 UTC`. So I think something like the following would work (this isn't tested code so it's maybe wrong):

```suggestion
        self.as_utc_duration() - Self::from_gregorian_utc_at_midnight(1970, 1, 1)
```

That way the as_utc_duration function does all of the leap second calculations. (It also may be worth making that as_utc_duration function `pub`).

src/epoch.rs
```diff
@@ -332,6 +332,35 @@ impl Epoch {
...
+        let cnt = e.get_num_leap_seconds();
+        // We have the time in TAI. But we were given UNIX (UTC).
+        // Hence, we need to _add_ the leap seconds  to get the actual TAI time.
+        // TAI = UNIX + leap_seconds + UNIX_OFFSET_UTC_SECONDS <=> UTC = TAI - leap_seconds - UNIX_OFFSET_UTC_SECONDS
+        e.0 += i64::from(cnt) * Unit::Second;
+        e
+    }
+
+    /// Initialize an Epoch from the provided UNIX milisecond timestamp since UTC midnight 1970 January 01.
+    pub fn from_unix_miliseconds(seconds: f64) -> Self {
```
ChristopherRabotin, line 350:
```suggestion
    pub fn from_unix_milliseconds(seconds: f64) -> Self {
```

src/epoch.rs
```diff
@@ -332,6 +332,35 @@ impl Epoch {
         }) + Unit::Second * SECONDS_GPS_TAI_OFFSET
     }
 
+    /// Initialize an Epoch from the provided UNIX second timestamp since UTC midnight 1970 January 01.
+    pub fn from_unix_seconds(seconds: f64) -> Self {
+        let mut e = Self::from_tai_duration(
```
ChristopherRabotin, line 337: I think this could be simplified as
```suggestion
        Self::from_utc_seconds(seconds + UNIX_OFFSET_UTC_SECONDS)
```

ChristopherRabotin (changes requested):
Look good, thanks! I'm proposed a few simplifications

src/epoch.rs
mkolopanis: Easy enough. I'll make sure I didn't add any more of these.

src/epoch.rs
mkolopanis: Interesting idea. I'll test this approach.

src/epoch.rs
mkolopanis: removed in 8377a00cff4f50377f570afcc0c002e9ba52d069

src/epoch.rs
mkolopanis: After your suggestion about using the `from_gregorian_utc_at_midnight` I found I could just get rid of the `UNIX_OFFSET_UTC_SECONDS` const entirely. Making an epoch from gregorian UTC then converting to a duration is equivalent and less numbers to memorize.

ChristopherRabotin:
Oh sorry, I totally forgot about it! Thanks for the reminder, I'll have a
look at it soon, hopefully in a few hours.

ChristopherRabotin (approved):

ChristopherRabotin:
Thanks!

PR #126, opened 2022-05-21
first version of STM32 benchmark

pwnorbitals:
For now, panics at 

```
23:24:52.475 Panic : PanicInfo { payload: Any { .. }, message: Some(assertion failed: `(left == right)`
23:24:52.475   left: `Duration { centuries: 1, nanoseconds: 2332740342346940416 }`,
23:24:52.475  right: `Duration { centuries: -32768, nanoseconds: 3155760000000000000 }`), location: Location { file: "src\\main.rs", line: 152, col: 13 }, can_unwind: true }
```

But all the building blocks are here. We might want to find a way to measure real elapsed time instead of cycles too, I still need to find a way to do that correctly.


ChristopherRabotin:
This looks really cool, although I don't quite understand everything.

It seems like the multiplication of a unit with a given value leads to a different duration if that value is in an f64 and in a i64. Is that right?

pwnorbitals:
Indeed, you're understanding this right. The bug is reproducible, here's a second output sample :

```
22:06:19.578 Panic : PanicInfo { payload: Any { .. }, message: Some(assertion failed: `(left == right)`
22:06:19.578   left: `Duration { centuries: -1, nanoseconds: 661696595250774016 }`,
22:06:19.578  right: `Duration { centuries: 32767, nanoseconds: 3155760000000000000 }`), location: Location { file: "src\\main.rs", line: 155, col: 13 }, can_unwind: true }
```

Currently working on the time measurement method, if you have any idea about the multiplication bug or any hint that could point me in the right direction I'd be glad ;)

ChristopherRabotin:
I have an idea! I think that two things happen:
1. The conversion from i64 to i64 does not return the same number
2. The numbers are actually really big so the duration multiplication overflows: with std enabled, this is caught. But I suspect without std, it probably does not work well.

Here is the code I used with std:

```rust
#[cfg(test)]
mod tests {
    use crate::{Errors, ParsingErrors, TimeSystem, Unit};
extern crate rand_core;
    #[test]
    fn stm32_test() {
        use self::rand_core::{OsRng, RngCore};

        const TEST_VALUES: u32 = 20;
        const ITERS_PER_TEST: u32 = 200;
        let mut rng = OsRng::default();

        for test_index in 0..TEST_VALUES {
            let mut buf1 = [0u8; core::mem::size_of::<i64>()];
            rng.fill_bytes(&mut buf1);
            let mut val1 = i64::from_be_bytes(buf1);
            let mut val2 = val1 as f64;
            println!("{val1} {val2}");

            let mut buf2 = [0u8; core::mem::size_of::<i64>()];
            rng.fill_bytes(&mut buf2);
            let mut val3 = i64::from_be_bytes(buf2);
            let mut val4 = val3 as f64;
            // println!("{val3} {val4}");

            for iter in 0..ITERS_PER_TEST {
                assert_eq!(Unit::Minute * val1, Unit::Minute * val2);
                assert_eq!(Unit::Second * val3, Unit::Second * val4);
            }
        }
    }
```

This immediately fails:
```
running 1 test
-2672162611803597408 -2672162611803597300
thread 'tests::stm32_test' panicked at 'attempt to multiply with overflow', src/duration.rs:524:29
stack backtrace:
   0: rust_begin_unwind
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/panicking.rs:143:14
   2: core::panicking::panic
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/panicking.rs:48:5
   3: <hifitime::duration::Unit as core::ops::arith::Mul<i64>>::mul
             at ./src/duration.rs:524:29
   4: hifitime::tests::stm32_test
             at ./src/lib.rs:228:28
   5: hifitime::tests::stm32_test::{{closure}}
             at ./src/lib.rs:207:5
   6: core::ops::function::FnOnce::call_once
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/ops/function.rs:227:5
   7: core::ops::function::FnOnce::call_once
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/ops/function.rs:227:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
test tests::stm32_test ... FAILED
```

Note how `val1` from i64 and `val2` as f64 are different numbers, `-2672162611803597408` and  `-2672162611803597300`.

It's also possible that the random numbers are really big because of the `OsRng` and that the behavior on the stm32 is quite different. In any case, I'll create an issue to add model verification to hifitime with [kani](https://model-checking.github.io/kani/getting-started.html). This should check that overflows don't happen or are prevented before they happen.

pwnorbitals:
Good catch. The random numbers aren't generated through OsRng but through the hardware random generator then transformed into the right number type via `from_le_bytes()`. Not sure how to fix yet, will think about it

pwnorbitals:
It seems to overflow only in one of the two cases, the i64. Is this expected ? Should we use some `checked` operations to avoid or constrain overflows ?

Also yes, f64 is losing a bit of precision at i64::MAX (1e19) as shown in a graph on IEEE754 Wiki page. Precision of one unit (1) is achieved at around 1e16, so maybe in the short term I should limit the range of generated i64 to guarantee i64 = f64.round() ?

ChristopherRabotin:
Yes, the multiplications should be checked and bounded to their max if an overflow occurs. I don't think that the operations should return a `Result` type though because that would make these operations just hard to use.

PR #133, opened 2022-08-02
derive Ord and Eq for Epoch

mkolopanis:
Derives Ord and Eq for epoch.
I also added a small test to check expected orderings between two epochs.


ChristopherRabotin:
Looks good, thanks!
For the std cmp ordering in the test, you can use core::cmp::ordering if the std feature is not available. That change would allow the tests to check the equality even without std.

mkolopanis:
Thanks for the tip, always forget about core counterparts

ChristopherRabotin:
Thanks for the PR! I'll release this with version 3.3 this weekend if that works for you too.

PR #143, opened 2022-09-13
Fix num-traits usage for no-std

simpsoneric:
When attempting to compile for a target: aarch64-unknown-none I was failing due to a couple reasons.
They both go back to there is no-std library reference available for that architecture.
Compiling with "no_std" on my linux target works fine.

If you attempt a build with cargo such as:

```
cargo build --no-default-features --target=aarch64-unknown-none -Z build-std=core
```
The current hifitime/master was failing.

Thought I'd pass along my current patches.
Feel free to edit, change, or disregard.

### Feature resolver issues

I believe this goes back to the resolver dependency as shown in:
- https://github.com/rust-num/num-traits/issues/138

Added resolver="2" to the cargo.toml to fix this.

Another option would be to upgrade to the "edition = 2021" in cargo.toml.
The 2021 edition defaults to the new resolver.

### no-std and various float math functions

Also, in order to use the `sin()` and some other functions, I needed to add back the `libm` num-traits dependency.
Those functions do not exist in the current core.

Maybe at some point in the future the [core intrinsics sinf64()](https://doc.rust-lang.org/core/intrinsics/fn.sinf64.html) related functions will be stabilized.

### DIFF
--- a/.github/workflows/formal_verification.yml
+++ b/.github/workflows/formal_verification.yml
@@ -4,7 +4,7 @@ on:
     branches:
       - master
 
-name: Test Workflow
+name: Formal Verification Workflow
 
 jobs:
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -45,10 +45,28 @@ jobs:
       
       - name: Test (no default features)
         run: cargo test --no-default-features
-      
+
       - name: Test (asn1der)
         run: cargo test --features asn1der
 
+  test_no_std:
+    name: no-std Nightly Test Suite
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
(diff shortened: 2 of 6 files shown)


ChristopherRabotin:
Thanks for finding and fixing that bug! As soon as the tests go through, I'll merge this in.

ChristopherRabotin:
Merged!

Note that I'm still doing some tests on version 3.4 before officially publishing it on crates.io, but I hope to be done with those tests within a week.

simpsoneric:
Sounds good, better to make sure it works than to rush out and break things :-)

ChristopherRabotin:
@simpsoneric Published! You should be able to add version `3.4.0` to your Cargo.toml. Let me know if you encounter any issues.

PR #152, opened 2022-10-09
Use lexical_core for parsing durations

### LINKED ISSUE #151: Duration parsing should allow for combination of units
`humantime` supports durations like `1hour 15min 5ns`, but that is not supported in `hifitime`. It should be.

This should also remove the dependency on `Regex`.

This issue should also allow conversion to `std::time::Duration` when `std` is enabled.

ChristopherRabotin:
Closes #151

Signed-off-by: Christopher Rabotin <christopher.rabotin@gmail.com>

### DIFF
--- a/.github/workflows/formal_verification.yml
+++ b/.github/workflows/formal_verification.yml
@@ -6,17 +6,21 @@ on:
 
 name: Formal Verification Workflow
 
+env:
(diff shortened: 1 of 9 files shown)


ChristopherRabotin:
Kani bug report filed: https://github.com/model-checking/kani/issues/1768 .

PR #160, opened 2022-10-16
optimise gregorian

conradludgate:
The loops that were originally there didn't optimise well. This new code optimises much better with less branches

Screenshot 2022-10-16 at 11 01 36

The one regression is a false positive and on re-runs switched between + and -

### DIFF
--- a/benches/bench_epoch.rs
+++ b/benches/bench_epoch.rs
@@ -17,7 +17,6 @@ pub fn criterion_benchmark(c: &mut Criterion) {
 
     c.bench_function("TT", |b| {
         b.iter(|| {
-            // TT is too slow now! Used to be 184ns, is now 241
             let e = Epoch::from_gregorian_utc_hms(2015, 2, 7, 11, 22, 33);
             e.as_tt_seconds();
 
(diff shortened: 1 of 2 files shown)


ChristopherRabotin:
Wow, this is a fantastic improvement in speed! Thanks!

I'll merge this is right away. Do you need this to be on crates.io right now or can it wait a bit?

conradludgate:
I'm in no rush, I saw this on reddit and thought I'd help out. I'm on a mission to make all time libraries efficient :)

ChristopherRabotin:
If there are any features you need that would make you consider using `hifitime` instead of another library at TrueLayer, let me know. My use case is quite narrowly focused on engineering and scientific applications, so I may miss the broader requirements that folks need, but I would really like it if hifitime became as useful to the Rust community as time.rs or chrono.

conradludgate:
I will definitely let you know. The kind of features we make use of are RFC3339 formatting, and recently working with dates only - eg adding months or finding the monday of week 12. I understand this use case is quite different to what hifitime is currently intending

PR #161, opened 2022-10-16
Remove redundant tests

SUPERCILEX:
Kani `any` is a symbolic value meaning all valid states will be solved for. See here: https://model-checking.github.io/kani/tool-comparison.html?highlight=exhau#comparison-with-other-tools

### DIFF
--- a/src/duration.rs
+++ b/src/duration.rs
@@ -1139,26 +1139,7 @@ fn deser_test() {
 
 #[cfg(kani)]
 #[kani::proof]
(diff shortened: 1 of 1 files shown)


ChristopherRabotin:
That's a good idea. Let's wait until Wednesday to run the tests: I want to check that the Kani report still shows the same level of coverage. The Kani team is releasing an update on Wednesday that will fix the `const fn` functions in the model checker.

PR #166, opened 2022-10-22
introducing GNSS timescales

gwbres:
GST: Galileo System time is defined as starting 13 seconds
prior midnight on August 22nd 1999
GPST: GPS System time is defined as starting on Jan 6 1980 midnight
BDT: BeiDou system time scale, defined as starting on Jan 1st 2006 midnight.

All of them do not take leap second into account.

### DIFF
--- a/src/epoch.rs
+++ b/src/epoch.rs
@@ -362,7 +362,7 @@ impl Epoch {
     #[must_use]
     /// Initialize an Epoch from the provided duration since January 1st midnight
     pub fn from_bdt_duration(duration: Duration) -> Self {
-        let mut me = Self::from_utc_duration(BDT_REF_EPOCH.to_utc_duration() + duration);
+        let mut me = Self::from_tai_duration(BDT_REF_EPOCH.to_tai_duration() + duration);
         me.time_scale = TimeScale::BDT;
         me
     }
@@ -1613,6 +1613,7 @@ impl Epoch {
     /// This will return an overflow error if more than one century has past since the reference epoch in the provided time scale.
     /// If this is _not_ an issue, you should use `epoch.to_duration_in_time_scale().to_parts()` to retrieve both the centuries and the nanoseconds
     /// in that century.
+    #[allow(clippy::wrong_self_convention)]
     fn to_nanoseconds_in_time_scale(&self, ts: TimeScale) -> Result<u64, Errors> {
         let (centuries, nanoseconds) = self.to_duration_in_time_scale(ts).to_parts();
         if centuries != 0 {
@@ -1906,7 +1907,7 @@ impl Epoch {
     #[must_use]
     /// Returns `Duration` past BDT (BeiDou) time Epoch.
     pub fn to_bdt_duration(&self) -> Duration {
-        self.to_utc_duration() - BDT_REF_EPOCH.to_utc_duration()
+        self.to_tai_duration() - BDT_REF_EPOCH.to_tai_duration()
     }
 
     #[must_use]
--- a/src/epoch.rs
+++ b/src/epoch.rs
@@ -1638,9 +1638,9 @@ impl Epoch {
             TimeScale::TT => self.to_tt_duration(),
             TimeScale::TDB => self.to_tdb_duration_since_j1900(),
             TimeScale::UTC => self.to_utc_duration(),
-            TimeScale::GPST => self.to_gpst_duration() - GPST_REF_EPOCH.to_tai_duration(),
+            TimeScale::GPST => self.to_gpst_duration() + GPST_REF_EPOCH.to_tai_duration(),
             TimeScale::GST => self.to_gst_duration() + GST_REF_EPOCH.to_tai_duration(),
-            TimeScale::BDT => self.to_bdt_duration() + BDT_REF_EPOCH.to_utc_duration(),
+            TimeScale::BDT => self.to_bdt_duration() + BDT_REF_EPOCH.to_tai_duration(),
         }
     }
 
--- a/src/epoch.rs
+++ b/src/epoch.rs
@@ -974,9 +974,6 @@ impl Epoch {
                     if idx != s.len() - 1 {
                         // We have some remaining characters, so let's parse those in the only formats we know.
                         ts = TimeScale::from_str(s[idx..].trim())?;
-                        if ts == TimeScale::GPST {
-                            print!("yo");
-                        }
                     }
                     break;
                 }
--- a/src/epoch.rs
+++ b/src/epoch.rs
@@ -590,7 +590,7 @@ impl Epoch {
     /// starting August 21st 1999 midnight (UTC)
     /// (cf. <https://gssc.esa.int/navipedia/index.php/Time_References_in_GNSS>)
     pub fn from_gst_days(days: f64) -> Self {
-        Self::from_duration(days * Unit::Nanosecond, TimeScale::GST)
+        Self::from_duration(days * Unit::Day, TimeScale::GST)
     }
 
     #[must_use]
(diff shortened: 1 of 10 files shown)


gwbres:
Hello

@larsnaesbye and I would like to introduce support for GNSS time scales

The following PR intends at defining GPS, GAL, BDS and GLO Time scales. Our RINEX crate could seriously make use of the conversion method this crate would then provide.

Current PR needs more documentation and dedicated test methods

tests/epoch.rs
```diff
@@ -825,6 +825,9 @@ fn test_format() {
                     TimeScale::TDB => format!("{epoch:e}"),
                     TimeScale::TT => format!("{epoch:X}"),
                     TimeScale::UTC => format!("{epoch}"),
+                    TimeScale::GPST => format!("{epoch}"),
```
ChristopherRabotin, line 828: I think this tests will fail because there is no shortcut to print in GPST, GST, or BDT, but the initialization from these epochs will set the time scale correctly. Let's see, maybe I'm wrong.

ChristopherRabotin (commented):
This is really useful work, thank you very much! Thanks for also adding the Python functions for those new time scales.

I just left a few comments and I'm happy to work on implementing those with you. As you said, I think that a few more tests would be warranted, most notably one that checks the initial epoch.

I'll have a look at your RINEX crate, I didn't know it existed but I might have use for it in Nyx.

src/epoch.rs
```diff
@@ -340,6 +345,30 @@ impl Epoch {
...
+    pub fn from_gpst_duration(duration: Duration) -> Self {
+        let mut e = Self::from_tai_duration(duration) + Unit::Second * SECONDS_GPS_TAI_OFFSET;
+        e.time_scale = TimeScale::GPST;
+        e
+    }
+    
+    #[must_use]
+    /// Initialize an Epoch from the provided duration since 1980 January 6 13 seconds prior midnight 
+    pub fn from_gst_duration(duration: Duration) -> Self {
+        let mut e = Self::from_tai_duration(duration) + Unit::Second + SECONDS_GST_TAI_OFFSET;
```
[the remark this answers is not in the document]
gwbres: That is correct, this is a typo

gwbres:
Hello @ChristopherRabotin, thank you for the helpful review. I fixed the obvious first issues.
I am in the process of putting new methods to the test, next contribution should be more meaningful.
See you

gwbres:
Hello @ChristopherRabotin,

I defined Galileo and BeiDou timescales properly, 
i moved these constants into the `TimeScale` definition.

I researched the TAI-UTC (=leap second) offset on the day these timescales started: +32 for GAL(1999) and +33 for BDS(2006). So I injected it like it was previously done for GPS.

I don't know why, i'm getting 
* a  3/4 of a day error when a GAL timestamp is converted back to UTC
* a 1/2 day error when a BDS timestamp is converted back to UTC. 

Would you care enough to run the test on your side?
I simply copied the previous `gpst` tests and adapted to the "'new" timescales.
You can focus with `cargo test galileo_time_scale` 
and `cargo test beidou_time_scale`

ChristopherRabotin:
Hi Guillaume,

Thanks for all of the updates and the info on the time differences you're
see. I can have a look in a few hours, but might end up reviewing it only
tomorrow.

Best regards,
Chris

ChristopherRabotin:
I'm looking into this now, and I actually think that the original computation of GPST is wrong (i.e. it is wrong on the `master` branch): the TAI offset is a round number of days, but J1900 TAI starts at noon whereas GPST starts at midnight. So there should be an extra twelve hours somewhere. That would explain the twelve hour difference you're seeing in the Beidou computation.

I'm making some changes now and will push to your branch (if I have permissions) once I'm done. I'm also making some "cosmetic" changes to avoid a breaking change: I'm placing the GPS offsets back in `lib.rs` in case users of the library were using those before.

ChristopherRabotin:
Okay, have a look and let me know if you spot any errors. I think that I've correctly defined the time scales.

I reverted a few of your changes that consolidated some UTC computations: of all of the time scales, only UTC follows leap seconds, so I think that abstracting the leap seconds additions in a generic function adds confusion. 

In these changes, Galileo is defined in _UT_ (not UTC), as per UN outer space https://www.unoosa.org/pdf/icg/2016/gst2016.pdf , and is 13 seconds before the equivalent UTC Gregorian date. Also, Navipedia claims that BDT is in-line with UTC, but that is incorrect: I'll contact Navipedia to get that fixed: UNOOSA says that it only accounted leap seconds until its reference epoch, and that it's a continuous time scale from there-on (https://www.unoosa.org/pdf/icg/2016/Beidou-Timescale2016.pdf).

gwbres:
Hello Christopher,

this looks good, I'll convert the toppic to open PR. Thank you so much your time!!

> and is 13 seconds before the equivalent UTC Gregorian date

So it is indeed starting 13 seconds prior midnight UTC 😄 

> says that it only accounted leap seconds until its reference epoch, and that it's a continuous time scale from there-on

those are very good sources of information, I must keep them somewhere

>I'm making some changes now and will push to your branch (if I have permissions) once I'm done. I'm also making some "cosmetic" changes to avoid a breaking change

having `REF_EPOCHs`  constants like we do now is much more sane, its the best way to describe the offset to J1900

>I reverted a few of your changes that consolidated some UTC computations: of all of the time scales, 
> only UTC follows leap seconds, so I think that abstracting the leap seconds additions in a generic function adds confusion.

I was thinking this would help introducing Glonass. It is said to take leap seconds into account, that would be the second case where that is true. But I suggest to keep it like this for the moment, and introduce Glonnass later in a separate topic

larsnaesbye:
> 
> I'll have a look at your RINEX crate, I didn't know it existed but I might have use for it in Nyx.

That would be great. The entire GNSS community is in need of a community-driven open source suite for the IGS formats, especially RINEX. Most other tools out there are either closed-source, outdated, or both.

src/epoch.rs
```diff
@@ -490,44 +553,98 @@ impl Epoch {
...
+    pub fn from_gst_seconds(seconds: f64) -> Self {
+        Self::from_duration(seconds * Unit::Second, TimeScale::GST)
+    }
+
+    #[must_use]
+    /// Initialize an Epoch from the number of days since the GST Time Epoch,
+    /// starting August 21st 1999 midnight (UTC)
+    /// (cf. <https://gssc.esa.int/navipedia/index.php/Time_References_in_GNSS>)
+    pub fn from_gst_days(days: f64) -> Self {
+        Self::from_duration(days * Unit::Nanosecond, TimeScale::GST)
```
ChristopherRabotin, line 593:
```suggestion
        Self::from_duration(days * Unit::Day, TimeScale::GST)
```

src/epoch.rs
```diff
@@ -1227,6 +1352,59 @@ impl Epoch {
...
+    /// starting on August 21st 1999 Midnight UTC,
+    /// (cf. <https://gssc.esa.int/navipedia/index.php/Time_References_in_GNSS>).
+    fn init_from_gst_days(days: f64) -> Self {
+        Self::from_gst_days(days)
+    }
+
+    #[cfg(feature = "python")]
+    #[staticmethod]
+    /// Initialize an Epoch from the number of nanoseconds since the Galileo Time Epoch,
+    /// starting on August 21st 1999 Midnight UTC,
```
ChristopherRabotin, line 1376:
```suggestion
    /// starting on August 21st 1999 Midnight UT,
```

src/epoch.rs
```diff
@@ -1227,6 +1352,59 @@ impl Epoch {
...
+    /// (cf. <https://gssc.esa.int/navipedia/index.php/Time_References_in_GNSS>).
+    fn init_from_gst_seconds(seconds: f64) -> Self {
+        Self::from_gst_seconds(seconds)
+    }
+
+    #[cfg(feature = "python")]
+    #[staticmethod]
+    /// Initialize an Epoch from the number of days since the Galileo Time Epoch,
+    /// starting on August 21st 1999 Midnight UTC,
```
ChristopherRabotin, line 1367:
```suggestion
    /// starting on August 21st 1999 Midnight UT,
```

src/epoch.rs
```diff
@@ -1227,6 +1352,59 @@ impl Epoch {
         Self::from_gpst_nanoseconds(nanoseconds)
     }
 
+    #[cfg(feature = "python")]
+    #[staticmethod]
+    /// Initialize an Epoch from the number of seconds since the Galileo Time Epoch,
+    /// starting on August 21st 1999 Midnight UTC,
```
ChristopherRabotin, line 1358:
```suggestion
    /// starting on August 21st 1999 Midnight UT,
```

ChristopherRabotin (approved):
Just a few more minor cleanups I'll do before merging this in and releasing it.

gwbres:
> Sorry, I had to go to work, so I'll release it tonight. It will be in release `3.6.0` because of the added time scales in `TimeScale` enum.

No hurries, there is no emergency regarding this topic. 
Work day almost over here, do you still work in Boulder CO ? I used to live there a couple of years ago

ChristopherRabotin:
Yes, I'm still in Colorado. You worked at NIST, is that right? I'm happy to chat by email christopher _dot_ rabotin _at_ gmail ;-)

PR #170, opened 2022-11-06
epoch: derive Hash

gwbres:
Hash is required if we want structures like "HashMap" or "BTreeMap"
indexed by an Epoch


ChristopherRabotin:
Thanks! That is definitely useful.

I'm not sure why Clippy doesn't like it. Here is the explanation: https://rust-lang.github.io/rust-clippy/master/index.html#derive_hash_xor_eq . Any idea how to fix this? Should `Hash` be implemented manually instead of derived? I want to keep the `PartialEq` and `Eq` "derivations" for backward compatibility (and it's a feature I use a lot).

gwbres:
Hello @ChristopherRabotin 

> Any idea how to fix this? Should `Hash` be implemented manually instead of derived? I want to keep the `PartialEq` and `Eq` "derivations" for backward compatibility (and it's a feature I use a lot).

https://www.philipdaniels.com/blog/2019/rust-equality-and-ordering/

My mere understanding is that Hash and Eq are closely related.
When implementing Eq yourself you can do whatever you want, like disregarding inner object of a structure (for instance, have `Eq` based only off `centuries` in Duration == Duration).
But auto derivation / auto implementation make use of all inner objects (I presume). 
So want must be very careful when mixing Auto and self implemented traits like this, especially these two

ChristopherRabotin:
Perfect, I think that implementing the hasher manually works well (and makes clippy happy). Thanks!

PR #175, opened 2022-11-15
epoch, duration: improve and fix serdes feature

gwbres:

### DIFF
--- a/src/duration.rs
+++ b/src/duration.rs
@@ -21,8 +21,8 @@ use core::fmt;
 use core::hash::{Hash, Hasher};
 use core::ops::{Add, AddAssign, Div, Mul, Neg, Sub, SubAssign};
 
-#[cfg(feature = "std")]
-use serde::{de, Deserialize, Deserializer};
+#[cfg(feature = "serde")]
+use serde_derive::{Deserialize, Serialize};
 
 use core::str::FromStr;
 
@@ -57,6 +57,7 @@ pub const NANOSECONDS_PER_CENTURY: u64 = DAYS_PER_CENTURY_U64 * NANOSECONDS_PER_
 #[derive(Clone, Copy, Debug, PartialOrd, Eq, Ord)]
 #[repr(C)]
 #[cfg_attr(feature = "python", pyclass)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct Duration {
     pub(crate) centuries: i16,
     pub(crate) nanoseconds: u64,
@@ -755,18 +756,6 @@ impl Duration {
     /// Create a new duration from the truncated nanoseconds (+/- 2927.1 years of duration)
     fn init_from_truncated_nanoseconds(nanos: i64) -> Self {
         Self::from_truncated_nanoseconds(nanos)
-    }
-}
-
-#[cfg(feature = "std")]
-#[cfg(not(kani))]
-impl<'de> Deserialize<'de> for Duration {
-    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
-    where
-        D: Deserializer<'de>,
-    {
-        let s = String::deserialize(deserializer)?;
-        FromStr::from_str(&s).map_err(de::Error::custom)
     }
 }
 
--- a/src/epoch.rs
+++ b/src/epoch.rs
@@ -28,8 +28,8 @@ use pyo3::prelude::*;
 #[cfg(feature = "python")]
 use pyo3::pyclass::CompareOp;
 
-#[cfg(feature = "std")]
-use serde::{de, Deserialize, Deserializer};
+#[cfg(feature = "serde")]
+use serde_derive::{Deserialize, Serialize};
 
 use core::str::FromStr;
 #[cfg(feature = "std")]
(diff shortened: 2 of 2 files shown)


gwbres:
Hello @ChristopherRabotin,

the serdes traits for `Duration` and `Epoch` do not work on my side.  
When I try to import these two in structures that need serdes traits, it complains the traits are not implemented.

In the previous code I don't understand why all of these options were depending on "std" instead of "serde".

`Timescale` does work, I pushed it myself a couple of weeks ago using the auto derivation.
I modified the previous code to use auto derivation. I don't think we need a custom implementation either, the two structures are very straightforward and only comprise native types

src/epoch.rs
```diff
@@ -159,6 +159,7 @@ const CUMULATIVE_DAYS_FOR_MONTH: [u16; 12] = {
 #[derive(Copy, Clone, Eq)]
 #[repr(C)]
 #[cfg_attr(feature = "python", pyclass)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
```
ChristopherRabotin, line 162:
```suggestion
#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
```

src/duration.rs
```diff
@@ -57,6 +57,7 @@ pub const NANOSECONDS_PER_CENTURY: u64 = DAYS_PER_CENTURY_U64 * NANOSECONDS_PER_
 #[derive(Clone, Copy, Debug, PartialOrd, Eq, Ord)]
 #[repr(C)]
 #[cfg_attr(feature = "python", pyclass)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
```
ChristopherRabotin, line 60:
```suggestion
#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
```



Yes! I think that was missing. Also, this issue should have been caught in the tests, but I know that the tests for serde are crappy ... actually did I remove the test? I don't see it right now!

src/duration.rs
```diff
@@ -21,8 +21,8 @@ use core::fmt;
 use core::hash::{Hash, Hasher};
 use core::ops::{Add, AddAssign, Div, Mul, Neg, Sub, SubAssign};
 
-#[cfg(feature = "std")]
-use serde::{de, Deserialize, Deserializer};
+#[cfg(feature = "serde")]
```
ChristopherRabotin, line 24:
```suggestion
#[cfg(feature = "std")]
```

There isn't any "serde" feature in the `Cargo.toml` so I'm guessing that this line causes the serde imports to never be executed.

From Cargo.toml:

```
[features]
default = ["std"]
std = ["serde", "serde_derive"]
asn1der = ["der"]
python = ["std", "asn1der", "pyo3"]
```

When the "std" feature is enabled, it should be importing serde, but I think that you've nailed the source of the issue with your change in Duration (I'll write another comment).

src/epoch.rs
```diff
@@ -28,8 +28,8 @@ use pyo3::prelude::*;
 #[cfg(feature = "python")]
 use pyo3::pyclass::CompareOp;
 
-#[cfg(feature = "std")]
-use serde::{de, Deserialize, Deserializer};
+#[cfg(feature = "serde")]
```
ChristopherRabotin, line 31:
```suggestion
#[cfg(feature = "std")]
```

ChristopherRabotin (approved):
Deriving serde is definitely nicer than the custom implementations I had. Could you add some tests for the new serialization because it seems like I removed those.

src/duration.rs
gwbres: Well I'll correct myself. "Serde::Serialize" and "Deseriliaze" and also auto derivations are imported with "std"
so your current Cargo.tml content.
It would actually work if you removed these imports from std too, with I explained just above this comment

ChristopherRabotin (approved):

.github/workflows/tests.yml
```diff
@@ -61,6 +61,9 @@ jobs:
 
       - name: Test (no default features)
         run: cargo test --no-default-features
+      
+      - name: Test (all features)
+        run: cargo test --all-features
```
ChristopherRabotin, line 66: The `python` feature can only be built with the PyO3 tools. So this line needs to be all of the features, minus the `python` feature.

gwbres: Apparently --no-default-features only tests "std" (that is my understanding)
Maybe we should create a dedicated step like "ANS1" ? I don't know

src/duration.rs
gwbres: Same thing here.. 

`rust_docs` clearly lacks some capabilities regarding crate features. They are totally undocumented in the API. Like there is absolutely no mean to know what is available besides reading the Cargo.toml..

There is a rust crate that allows documenting a crate feature, but its documentation must lie in Cargo.toml, so it will grow a little too much. And it's also one more dependency required for your crate. I think they should come with a solution to this problem in a near future

.github/workflows/tests.yml
ChristopherRabotin: Yes, that would also work

src/epoch.rs
gwbres: so in conclusion, "serde" already exist in the current Cargo.toml
and it's usually invoked with "serde"

PR #180, opened 2022-12-02
timeofweek: introduce time of week construction and conversion

gwbres:
* (wk, tow) is usually how an epoch is described by a GNSS receiver, usually in a GNSS timescale.
`wk` is a rolling counter, counting the elapsed weeks since the first epoch of that time scale.
`tow` is the amount of seconds since Sunday midnight of that week.

`tow` is usually expressed in [s] or [ms] by receivers, we use [ns] in our case, to maintain crate consistency and never be limited.


gwbres:
Hello @ChristopherRabotin,

I'd like to introduce these two methods to deal with GNSS receivers easily. 
`from_timeofweek`  would help parse an epoch description from a GNSS receiver very easily.
`to_timeofweek` would help perform computations from GNSS receiver data.

The `wk` week number counter in a given GNSS time scale works fine, it's fairly easy.
But the `tow` would be the number of nanoseconds, in that week, considering Sunday midnight as the starting point.
We need some kind of logic to determine to which week day, the previously (and correctly) evluated `wk` corresponds to

ChristopherRabotin:
Sure, that looks like it could be helpful, so thanks for starting this work.
I'm not sure I understand: is `time_of_week` the duration since the week started? If so, when does the week start? On midnight between Saturday and Sunday, or on midnight between between Sunday and Monday? I contribute to this PR over the weekend.

gwbres:
Hello Chris, I improved the PR description and tried to improve the autodocs. 

> I'm not sure I understand: is `time_of_week` the duration since the week started? 

maybe `to_timeofweek` is poorly named.

For example, GNSS vehicles describe an epoch in their timescale, with a rolling week counter ("wk" - maybe we should also improve some variables name), and "tow" is the offset between the current instant and the closest sunday midnight.
See [this UTC/GPST calculator](https://www.labsat.co.uk/index.php/en/gps-time-calculator)

I managed to create the construction, currently named `from_timeofweek (wk, tow)`  which seems to work fine, and the existing objects fill the requirements.

For the reverse operation `(wk , tow_offset) = epoch.to_timeofweek()` it is more tricky because Hifitime does not have weekday and week determination methods. But it was surprisingly easier than I thought.

See this new method

```rust
pub fn Epoch::weekday_tai(&self) -> u8 {}
```

which gives an unsigned integer number wrapping at 6, which will help describe a weekday, starting on 0: Monday

Then I introduced 

```rust
pub fn closest_utc_start_of_week(&self) -> Self 
```

which for Self, returns the closest Sunday Midnight UTC. epoch   
In this one, I struggle a bit to strip the H:M:S offset, and bring it back to midnight.

This is the main information I need to evaluate what I notated `tow` so far: the offset in the current week.

in 

```rust 
pub fn Epoch::to_timeofweek_utc(&self) -> (u32, u64)
```

the rolling week counter is easily determined.
Then I determine the closest sunday midnight
I struggle a bit to determine the elapsed duration since Self and that day.
I don't understand why, in our implementation we have 37 leap seconds, while the calculator above says it should be 18

ChristopherRabotin:
Understood, thanks Guillaume! I'll start the review soon and hope to be able to help on the bug this weekend. This looks very helpful even to other issues like #183 .

gwbres:
> Understood, thanks Guillaume! I'll start the review soon and hope to be able to help on the bug this weekend. This looks very helpful even to other issues like #183 .

Yep! indeed. I decided not to introduce a new Unit or timeserie, to keep things simple. But a Week granularity would help.
Anyways, we can even split this PR, get rid of the GNSS stuff and keep the `WeekDay` counter method, which works great and will be useful in itself

src/epoch.rs
```diff
@@ -2226,6 +2226,82 @@ impl Epoch {
...
+            TimeScale::GST => self.to_gst_days(),
+            TimeScale::BDT => self.to_bdt_days(),
+            TimeScale::TT => self.to_tt_days(),
+            TimeScale::TAI => self.to_tai_days(),
+            TimeScale::UTC => self.to_utc_days(),
+            TimeScale::TDB => self.to_tdb_seconds() / SECONDS_PER_DAY,
+            TimeScale::ET => self.to_et_seconds() / SECONDS_PER_DAY,
+        };
+        // wk: rolling week counter into timescale
+        let wk = (days / 7.0) as u32;
```
ChristopherRabotin, line 2263: Should this be `floor()`?

Also, just to make sure I understand, the reason this works is because the reference epoch is on 01 January 1900 TAI and that day happens to be a Monday, is that right?

src/epoch.rs
```diff
@@ -2226,6 +2226,82 @@ impl Epoch {
...
+        // tow: number of nanoseconds between self and closest sunday midnight / start of week
+        let start_of_week = self.closest_utc_start_of_week();
+        let dw = *self - start_of_week; // difference in weekdays [0..6]
+        let (_, d, h, m, s, ms, us, ns) = dw.decompose();
+        let tow = Duration::from_days(d as f64)
```
ChristopherRabotin, line 2268: You can use `Duration::compose(...)` here with the same parameters that `decompose` returns. Although I don't quite understand why you call `decompose` if you rebuild the Duration with all of the elements from `decompose` just afterwards. It seems like you could just do:

```suggestion
        let (_, mut tow) = dw.to_parts();
```

But maybe I'm misunderstanding something.

src/epoch.rs
```diff
@@ -2226,6 +2226,82 @@ impl Epoch {
...
+            + Duration::from_microseconds(us as f64)
+            + Duration::from_nanoseconds(ns as f64);
+        let (_, mut tow) = tow.to_parts();
+        tow += self.leap_seconds(true).unwrap_or(0.0) as u64 * 1_000_000_000;
+        (wk, tow)
+    }
+
+    /// Returns weekday counter in TAI timescale,
+    /// 0: Monday, ..., 6: Sunday
+    pub fn weekday_tai(&self) -> u8 {
```
ChristopherRabotin, line 2282:
```suggestion
    pub fn weekday_tai(&self) -> Weekday {
```

And then create a new file called `weekday.rs` with the enum:

```rust
#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
pub enum Weekday {
    Monday,
    Tuesday,
    Wednesday,
    Thursday,
    Friday,
    Saturday,
    Sunday,
}
```

src/epoch.rs
```diff
@@ -2226,6 +2226,82 @@ impl Epoch {
...
+        // we're helped here, because J1900 was a monday :)
+        (self.to_tai_days() as u64).rem_euclid(7) as u8
+    }
+
+    /// Returns weekday counter in UTC timescale,
+    /// 0: Monday, ..., 6: Sunday
+    pub fn weekday_utc(&self) -> u8 {
+        Self::from_tai_duration(
```
ChristopherRabotin, line 2290: Does the following work?

```suggestion
        self.to_utc_duration().weekday_tai()
```

ChristopherRabotin (commented):
Just a few comments for when I can contribute back to this branch. This is great works, thanks Guillaume!

src/epoch.rs
```diff
@@ -2226,6 +2226,82 @@ impl Epoch {
...
+    pub fn to_timeofweek_utc(&self) -> (u32, u64) {
+        // fractionnal days in this time scale
+        let days = match self.time_scale {
+            TimeScale::GPST => self.to_gpst_days(),
+            TimeScale::GST => self.to_gst_days(),
+            TimeScale::BDT => self.to_bdt_days(),
+            TimeScale::TT => self.to_tt_days(),
+            TimeScale::TAI => self.to_tai_days(),
+            TimeScale::UTC => self.to_utc_days(),
+            TimeScale::TDB => self.to_tdb_seconds() / SECONDS_PER_DAY,
```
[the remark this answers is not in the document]
gwbres: > The same should be done for the ET conversion to make use of the Duration unit conversion (which is more precise).

correct

src/epoch.rs
gwbres: > Should this be `floor()`?

that is true

> Also, just to make sure I understand, the reason this works is because the reference epoch is on 01 January 1900 TAI and that day happens to be a Monday, is that right?

that's correct

src/epoch.rs
gwbres: This makes things already much simpler.

src/epoch.rs
gwbres: for this, we have to introduce a  `Unit::Week` and make  `Duration` knowledgeable

src/epoch.rs
```diff
@@ -2226,6 +2226,82 @@ impl Epoch {
         me
     }
 
+    /// Builds an Epoch from given `wk` week counter into the desired Time scale.
+    /// `ns` is the amount of nanoseconds into that week from closest Sunday Midnight.
+    /// This is usually how GNSS receivers describe a GNSS time scale epoch
+    #[cfg(feature = "python")]
+    #[staticmethod]
+    pub fn from_timeofweek(wk: u32, ns: u64, ts: TimeScale) -> Self {
```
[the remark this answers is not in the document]
gwbres: I think I solved this one, but I'll leave it open because i'm not 100% sure

ChristopherRabotin:
Okay, I've started the review, and I'll push to your branch in a bit. This is quite good work, thanks!

ChristopherRabotin:
Okay, I fixed the time of week calculation. I think it needs more testing, most notably reciprocity testing. Otherwise, this looks good!

ChristopherRabotin:
I don't think `from_time_of_week_utc` can work because `to_time_of_week` lost the information on the year, month, and the day. I'll see if I can think of a signature change and let you choose whether or not that is acceptable for your use-case.

tests/weekday.rs
```diff
@@ -0,0 +1,259 @@
...
+        } else {
+            assert_eq!(neg_delta, i64::from(day_num % 7 - 2) * Unit::Day);
+        }
+        // Test FromStr
+        assert_eq!(Weekday::from_str(&format!("{day}")).unwrap(), day);
+    }
+}
+
+#[test]
+fn test_next() {
```
gwbres, line 76: interesting approach with `next` and `previous`

src/epoch.rs
```diff
@@ -2226,6 +2241,193 @@ impl Epoch {
...
+        self.to_duration() - start_of_year.to_duration()
+    }
+
+    /// Returns the number of days since the start of the year.
+    pub fn day_of_year(&self) -> f64 {
+        self.duration_in_year().to_unit(Unit::Day)
+    }
+
+    /// Returns the hours of the Gregorian representation  of this epoch in the time scale it was initialized in.
+    pub fn hours(&self) -> u64 {
```
gwbres, line 2370: these will prove useful too

src/weekday.rs
```diff
@@ -0,0 +1,170 @@
...
+    fn from(u: u8) -> Self {
+        match u.rem_euclid(Self::MAX) {
+            0 => Self::Monday,
+            1 => Self::Tuesday,
+            2 => Self::Wednesday,
+            3 => Self::Thursday,
+            4 => Self::Friday,
+            5 => Self::Saturday,
+            6 => Self::Sunday,
+            _ => Self::default(), // Defaults back to default for other values.
```
gwbres, line 59: this default is probably a better idea

src/epoch.rs
```diff
@@ -2226,6 +2242,218 @@ impl Epoch {
...
+                    start_of_week = GST_REF_EPOCH;
+                }
+            }
+            TimeScale::TAI => {
+                if start_of_week < J1900_REF_EPOCH {
+                    start_of_week = J1900_REF_EPOCH;
+                }
+            }
+            _ => {}
+        }
```
gwbres, line 2278: @ChristopherRabotin,
the previous implementation behaved funny for extreme values (inside the first week of a given timescale).
Because a timescale possibly did not start on a monday, and `previous_weekday_at_midnight` would determine a day prior `REF_EPOCH` . It makes more sense to restrict to the timescale, since these rolling counters are usually implemented in the form of unsigned integer. `u16` is truly how a GPS vehicle count the elapsed weeks for instance.

With this, i'm able to get the tests to pass for extreme values, see the testbench.

Also, i don't unterstand why the UNIX timescale is not wrapped in our timescales, it could be convenient for it to benefit all the existing methods, not only these ones. 

I will try to get the reciprocal test to work today

src/epoch.rs
```diff
@@ -2226,6 +2242,218 @@ impl Epoch {
...
+    pub fn to_time_of_week(&self) -> (u32, u64) {
+        // fractional days in this time scale
+        let days = self.to_duration().to_unit(Unit::Day);
+        // wk: rolling week counter into timescale
+        let wk = (days / Weekday::DAYS_PER_WEEK).floor() as u32;
+        // tow: number of nanoseconds between self and previous sunday midnight / start of week
+        let mut start_of_week = self.previous_weekday_at_midnight(Weekday::Sunday);
+        // restrict start of week/sunday to the time scale starting day
+        // std::cmp::min( start_of_week, self.time_scale.ref_epoch)
+        match self.time_scale {
```
gwbres, line 2256: @ChristopherRabotin,

the current implementation behaves funny for extreme values, i'm interested in to_time_of_week for Keplerian equations solving, but it's actually all `Epoch::previous` and `Epoch::next` that are impacted. For these two, extreme values are simply within the first week of a given timescale.

`previous` or `next` have no constraint mechanism to make sure the returned Sunday is not prior  `REF_EPOCH`. I think it is something we should implement, and my current tweak should be moved to these methods, so all calculations requiring weekday operations also benefit from it.

The method i'm interested in is just a use case of it. I decided to return `u16`  but that is arbitrary, i did this because I know that is how GPS vehicles encode such information, so the end user does not have any cast operation to perform. But that is not the point here, all we need to target is correct arithmetic

src/epoch.rs
```diff
@@ -2226,6 +2242,218 @@ impl Epoch {
...
+    /// assert_eq!(epoch.previous(Weekday::Friday), Epoch::from_gregorian_utc_at_midnight(1988, 1, 1));
+    /// assert_eq!(epoch.previous(Weekday::Thursday), Epoch::from_gregorian_utc_at_midnight(1987, 12, 31));
+    /// assert_eq!(epoch.previous(Weekday::Wednesday), Epoch::from_gregorian_utc_at_midnight(1987, 12, 30));
+    /// assert_eq!(epoch.previous(Weekday::Tuesday), Epoch::from_gregorian_utc_at_midnight(1987, 12, 29));
+    /// assert_eq!(epoch.previous(Weekday::Monday), Epoch::from_gregorian_utc_at_midnight(1987, 12, 28));
+    /// assert_eq!(epoch.previous(Weekday::Sunday), Epoch::from_gregorian_utc_at_midnight(1987, 12, 27));
+    /// assert_eq!(epoch.previous(Weekday::Saturday), Epoch::from_gregorian_utc_at_midnight(1987, 12, 26));
+    /// ```
+    pub fn previous(&self, weekday: Weekday) -> Self {
+        let delta_days = weekday - self.weekday();
```
gwbres, line 2367: @ChristopherRabotin,

the current implementation behaves funny for extreme values, ie., early dates into a timescale, because we might return a Sunday for example in the use case i'm interested in, that is prior `REF_EPOCH`.

I think we should constraint the values to at worst REF_EPOCH here. It is sane for a timescale to not wrap prior t(0) which is correctly defined. But it depends what you prefer. And I think, but i might be wrong, `Duration` wight negative centuries is already implemented. But also, Duration and Epoch are two different things.

src/timescale.rs
```diff
@@ -99,6 +105,15 @@ impl TimeScale {
     pub const fn is_gnss(&self) -> bool {
         matches!(self, Self::GPST | Self::GST | Self::BDT)
     }
+    /// Returns Reference Epoch (t(0)) for given timescale
+    pub(crate) const fn ref_epoch(&self) -> Epoch {
+        match self {
+            Self::GPST => GPST_REF_EPOCH,
+            Self::GST => GST_REF_EPOCH,
+            Self::BDT => BDT_REF_EPOCH,
+            _ => J1900_REF_EPOCH,
```
gwbres, line 114:
```rust
_ => J1900_REF_EPOCH
```

is that true ?

tests/epoch.rs
```diff
@@ -1337,3 +1343,207 @@ fn test_minmax() {
...
+    // GPST
+    // https://www.labsat.co.uk/index.php/en/gps-time-calculator
+    // 01/12/2022 00:00:00 <=> (2238, 345_618_000_000_000)
+    //      2238 weeks since 1980 + 345_600_000_000_000 ns since previous Sunday
+    //                            +      18_000_000_000 ns for elapsed leap seconds
+    let epoch = Epoch::from_time_of_week(2238, 345_618_000_000_000, TimeScale::GPST);
+    assert_eq!(epoch.to_gregorian_utc(), (2022, 12, 01, 00, 00, 00, 00));
+    assert_eq!(epoch.to_time_of_week(), (2238, 345_618_000_000_000));
+
+    let epoch_utc = epoch.in_time_scale(TimeScale::UTC);
```
gwbres, line 1476: Hello Chris,
I just pushed more tests, especially using other TimeScales it works great.

But this test right here: switching to UTC and asking for a reciprocal, always ends up in a perfect +1day offset. %H:%M:%S:%ns fields are correct. 
Starting from a different timescale than GPST, GST or BDT for example, produce the same effect.

tests/weekday.rs
ChristopherRabotin: I think it can be useful as a general use-case, for setting cookies or whatever.

src/weekday.rs
ChristopherRabotin: Yes! I agree! I didn't want to change it because you had it as Sunday, but I'm happy to set it to Monday (it makes more sense to me). It's also the most common approach around the world: https://en.wikipedia.org/wiki/Week#/media/File:First_Day_of_Week_World_Map.svg .

src/timescale.rs
ChristopherRabotin: Yes, I'm pretty sure. To check, we can change the duration calculations for these time scales to use this `ref_epoch` const function: if the tests fail, then something is wrong.

Ah, actually, I think that ET and TDB need their own ref epoch of J2000 in their respective time scale.

tests/epoch.rs
ChristopherRabotin: Okay, I'll investigate. This might be a sign of an error when changing time scales, I'm not sure.

src/timescale.rs
gwbres: > Ah, actually, I think that ET and TDB need their own ref epoch of J2000 in their respective time scale

Yep, my concern was about these two, which I am not familiar with, and I saw you quite often refer to J2000 when dealing with them

gwbres:
Hello Chris,
in case you're interested in, [this is my use case](https://github.com/gwbres/rinex/pull/62/commits/4a60edabb7c4b43d535f52273e6dcb3cb5210553#diff-aa61ceb86d2a830013d4f6a03324dd3c7e2bebec54893beea348b8c607ae25b5R668).

To determine the position of a satellite vehicle in the sky, you need to solve Kepler's equations. 
Time at epoch k (`t(k)`) is one out of several parameters.
It is expressed in the timescale (so referenced to `GPST_REF_EPOCH`) and the vehicles stream the number of weeks since that day, and the amount of nanosecond within that week

ChristopherRabotin:
I cannot seem to figure out where this time of week bug with UTC time scales comes from. I'm pretty sure that something is wrong in the calculation of `time_of_week` due to some rounding somewhere, but I cannot understand where exactly. I'll flag this as a bug for now.

ChristopherRabotin:
I ended up implementing a hot fix. Sorry for the delay in getting this merged.

PR #188, opened 2022-12-06
src/epoch.rs: fix gh 187

gwbres:
- fix conversion from and to elapsed weeks

### DIFF
--- a/src/epoch.rs
+++ b/src/epoch.rs
@@ -1470,13 +1470,13 @@ impl Epoch {
         nanos: u32,
         time_scale: TimeScale,
     ) -> Self {
-        Self::from_gregorian(year, month, day, hour, minute, second, nanos, ts)
+        Self::from_gregorian(year, month, day, hour, minute, second, nanos, time_scale)
     }
 
     #[cfg(feature = "python")]
     #[staticmethod]
     fn init_from_gregorian_at_noon(year: i32, month: u8, day: u8, time_scale: TimeScale) -> Self {
-        Self::from_gregorian_at_noon(year, month, day, ts)
+        Self::from_gregorian_at_noon(year, month, day, time_scale)
     }
 
     #[cfg(feature = "python")]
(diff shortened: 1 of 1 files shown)


gwbres:
Hello Chris,

with the following `src/epoch.rs` modification, I can unlock reciprocal tests of every `time_of_week` existing tests.

gwbres:
> In general, I try to avoid operations on i128 because that's a very big
> type and will be very slow on embedded systems (most are 32 bit, so it
> needs four registers per operation).

Yeah I know, that is why I was trying something like this

```rust
2253     #[must_use]
2254     /// Converts this epoch into the time of week, represented as a rolling week counter into that time scale
2255     /// and the number of nanoseconds elapsed in current week (since closest Sunday midnight).
2256     /// This is usually how GNSS receivers describe a timestamp.
2257     pub fn to_time_of_week(&self) -> (u32, u64) {
2258         let days = self.to_duration().to_unit(Unit::Day);
2260         // nb of weeks
2261         let weeks = days / Weekday::DAYS_PER_WEEK;
2262         // nanoseconds within current week
2263         //   is total number of nanoseconds
2264         //   - number of nanoseconds contained in whole days
2264         let nanoseconds = self.to_duration().to_unit(Unit::Nanosecond) //total
2265             - weeks.fract() /7.0 /24.0 /3600.0 /1.0E9; // remove those already contained in whole "weeks"
2267         (weeks.trunc() as u32, nanoseconds as u64)
2268     }

File: src/timescale.rs [hunk 1]
```diff
@@ -121,6 +121,14 @@ impl TimeScale {
             Self::TT | Self::TAI | Self::UTC => J1900_REF_EPOCH,
         }
     }
+
+    pub const fn ref_weekday(&self) -> Weekday {
```
gwbres, line 125: I'll remove this,
since the current proposal does not use it

ChristopherRabotin (approved):

Issue #182, opened 2022-12-02
Support day of year initialization and formatting


ChristopherRabotin:
ISO8601 allows epochs to be initialized from the day in the year. This should also be supported in Hifitime. The initialization should be trivial to implement, but the formatting may be a notch more complicated.

The formatting should be performed via its own method, independently of `fmt::Display` or one of the standard tokens since all of these are used already.

Hence, it's likely that a new formatting struct be needed to store the desired format.

- [x] Add `from_day_of_year` and `year_days_of_year`
- [x] Build custom formatting structure

Issue #181, opened 2022-12-02
Support formatting from the 1989 C standard (like strptime and strftime)


ChristopherRabotin:
This would allow formatting date times with the typical tokens like `%Y-%m-%d`. Chrono supports this, and its support should be used as a template for the Hifitime implementation: https://docs.rs/chrono/latest/chrono/format/index.html , https://docs.rs/chrono/latest/chrono/format/strftime/index.html .

Time-rs also has extensive support for formatting: https://time-rs.github.io/book/api/well-known-format-descriptions.html .

ChristopherRabotin:
This formatting should heavily reuse the work in `parser.rs` as it already includes a number of tokens. 

I think that the `ending_char` should be set by the user. 

The formatting structure should also use `fmt::Result` in order to be compatible with no-std.

The format string could be internally represented as:
```rust
struct EpochFormat {
    items: [Option<EpochItem>; 16],
    num_items: usize
}

struct EpochItem {
    token: Token,
    sep_char: Option<char>,
    sep_char_alt: Option<char>,
}
```

Then, some formats could be defined at compile time. Date "only" would be (would be initialized at midnight):

```rust
let date_only = EpochFormat {
    items: [EpochItem { token: Token::Year, sep_char: Some('-'), sep_char_alt: None },
            EpochItem { token: Token::Month, sep_char: Some('-'), sep_char_alt: None },
            EpochItem { token: Token::Day, sep_char: Some('-'), sep_char_alt: None }],
    num_items: 3
}
```

PR #190, opened 2022-12-08
Support strftime and strptime equivalent for formatting and parsing

ChristopherRabotin:
For #182, this adds support for the computation of Day of Year (e.g. `059` for the 29th of February 2000). This also supports initialization of an Epoch provided the year, the Day of Year, and the time scale.

For #181, this adds support for custom formatting rules, as well as providing a number of standard formats like RFC2820. This also introduced the month name structure. This structure is used strictly for formatting and has no other use. It is therefore extremely limited in scope.

Notes on the implementation of the parsing: if the week day is provided in the format and the string to parse leads to an Epoch whose week day is different from the one provided, the parsing will return an error.

@gwbres I've added you as a reviewer since you've contributed a lot recently. I'll be self-reviewing the changes proposed here anyway, but a second pair of eyes is always useful.

Remaining TODOs:
- [x] Parsing provided an EpochFormat and a string
- [x] Convert the current `from_str` to the parsing method above (I'm unsure about this because the current `from_str` is quite well optimized so there may be some performance degradation if the switch is made).


gwbres:
Hello Chris,

> @gwbres I've added you as a reviewer since you've contributed a lot recently. I'll be self-reviewing the changes proposed here anyway, but a second pair of eyes is always useful.

thanks! I'll test this new content when I have a little time

Issue #184, opened 2022-12-02
Greatly improve formal verification testing


ChristopherRabotin:
Now that the HTML Kani report is available in the CI pipeline, it's quite easy to see what is covered and what isn't. The purpose of this issue is to increase the parts of the code that are formally verified.

- [x] Duration formal verification
    - [x] Verify reciprocity of constructors with their associated functions (like `from_truncated_nanoseconds` should match the result of `try_truncated_nanoseconds`)
    - [x] `to_seconds` shall be tested with very high durations (and will likely exhibit the rounding errors of `f64`s)
    - [x] `decompose` shall be tested with very high durations (there are expectations in the code but these aren't checked)
    - [x] `min` and `max` on durations should be tested to be correct (this will require the need for the kani assumptions)
- [x] Epoch formal verification (this is new)
    - [x] Build an `Duration::any()` function for kani (take inspiration on nalgebra if needed)
    - [x] For all time scales that support an initialization, ensure that `Epoch::from_$ts$_duration(Duration::any())` returns the same `Duration::any()`. Skip the ET time scale because it uses a newton Raphson scheme for building durations and that may introduce an error of up to 20 nanoseconds (IIRC).
    - [x] From all time scales that support initialization from seconds or days, repeat the test above.

ChristopherRabotin:
The `to_seconds` test takes over two hours on my machine, and even then I killed it. So we'll figure this out later.

PR #192, opened 2022-12-15
Extend formal verification

ChristopherRabotin:
Refer to #184 for progress.

Implements #184 

List of bugs fixed thanks to Kani:

+ `normalize()` in Duration (which is called after _every_ operation on a Duration) could overflow near the maximum or minimum number of centuries when the nanoseconds were still within the MIN/MAX of the structure.
+ `normalize()` in Duration would overflow when the duration was _near_ (but below) the MAX
+ `Duration::neg()` would overflow near MIN/MAX
+ `PartialEq` on `Duration` would overflow near MIN/MAX
+ `try_truncated_nanoseconds` would overflow on any duration where the input was larger than the MAX
+ `Ops::Add` would fail when the centuries was the `i16::MIN` but the nanoseconds were still valid
+ `Ops::Add` would overflow when the right hand side input was not normalized
+ `Ops::Sub` would underflow when subtracting a duration whose nanoseconds was greater than `self` and when `self.centuries` was near the MIN.

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,6 +284,14 @@ In order to provide full interoperability with NAIF, hifitime uses the NAIF algo
 
 # Changelog
 
+## 3.8.0
+Thanks again to [@gwbres](https://github.com/gwbres) for his work in this release!
+
++ Fix CI of the formal verification and upload artifacts, cf. [#179](https://github.com/nyx-space/hifitime/pull/179)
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -284,6 +284,14 @@ In order to provide full interoperability with NAIF, hifitime uses the NAIF algo
 
 # Changelog
 
+## 3.8.0
```
ChristopherRabotin, line 287: This is a work in progress, but I'm adding this to not forget the release notes.

ChristopherRabotin:
Woohoo! All of the formal verification passes! Will merge now.

[kani-formal-verification-reports.tar.gz](https://github.com/nyx-space/hifitime/files/10317854/kani-formal-verification-reports.tar.gz)

PR #208, opened 2023-03-14
Use `-Z sparse-registry` when running Kani

karkhaz:
This commit adds `-Z sparse-registry` to the cargo kani invocation, which is needed to update crates over HTTP.

This commit fixes model-checking/kani-github-action#30


karkhaz:
Closing this, the `sparse-registry` feature is already merged into stable and I don't think this would help.

Issue #210, opened 2023-03-18
Add support for QZSS and IRNSS time scale


ljbade:
The two time scales from RINEX 4.0 used in https://github.com/gwbres/rinex/ that is not supported by `TimeScale` is the QZSS and IRNSS System Time. I would like to make a feature request to add it.

The QZSS System Time has the same offset from TAI as the GPS Time Scale, however technically it is its own time scale as they have their own set of master clocks. Per the QZSS ICD https://qzss.go.jp/en/technical/download/pdf/ps-is-qzss/is-qzss-pnt-004.pdf

> 5.1. Time System
> The PNT time system shall is the QZSS time system (QZSST) shown below.
> (1) Definitions
> (a) Length of 1 second
> The length of 1 second in QZSST shall be identical to the International Atomic Time (TAI).
> (b) Offset between QZSST and TAI
> QZSST shall be delayed from TAI by 19 seconds.
> (c) Starting point of week number for QZSST
> The starting point of the week number for QZSST shall be the same as the GPS time system
> (GPST), which is 0:00 am (UTC) on January 6, 1980.
> (2) Navigation message reference time
> The parameters relating to time such as the SV clock parameter, mean motion and UTC
> parameters shall all be based on QZSST.

The IRNSS System Time has a 13 second offset from UTC. Per the IRNSS ICD https://www.isro.gov.in/media_isro/pdf/Publications/Vispdf/Pdf2017/irnss_sps_icd_version1.1-2017.pdf

> 5.7 IRNSS SYSTEM TIME
> The IRNSS system time is given as 27-bit binary number composed of two parameters as
> follows:
> The Week Number is an integer counter that gives the sequential week number from the
> origin of the IRNSS time. This parameter is coded on 10 bits appearing in the first subframe,
> which covers 1024 weeks (about 19 years).
> The Time of Week Count (TOWC) indicates the number of 12 second counts at which the
> next subframe will begin. It is represented in 17 bits. The TOW count value ranges from 1 to
> 50400 to cover one entire week. The Time Of Week (TOW) in seconds is obtained by
> multiplying TOWC with 12. The TOWC will have a value of 1 at 00:00:00 Sunday
> (Changeover from Saturday to Sunday).
> The IRNSS System Time start epoch shall be 00:00 UT on Sunday August 22 nd 1999
> (midnight between August 21 st and 22 nd ). At the start epoch, IRNSS System Time shall be
[...]

PR #228, opened 2023-05-20
support qzss time scale

gwbres:
Introduces QZSS time scale, which is straight forward in the sense it behaves like GPST strictly, but comes from a set of dedicated clocks.

Partial answer to issue #210

### DIFF
--- a/src/timescale.rs
+++ b/src/timescale.rs
@@ -200,7 +200,7 @@ impl From<TimeScale> for u8 {
 }
 
 /// Allows conversion of a u8 into a TimeSystem.
-/// Mapping: 1: TT; 2: ET; 3: TDB; 4: UTC; 5: GPST; 6: GST; 7: BDT; anything else: TAI
+/// Mapping: 1: TT; 2: ET; 3: TDB; 4: UTC; 5: GPST; 6: GST; 7: BDT; 8: QZSST; anything else: TAI
 impl From<u8> for TimeScale {
     fn from(val: u8) -> Self {
         match val {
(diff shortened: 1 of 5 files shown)


ChristopherRabotin:
Woohoo ! Great, thanks, I plan to pick up hifitime this week so I'll have a look soon. I changed the merge branch to `4.0.0-dev` which will be the development branch for version 4.0.0. There are about 400 downloads of hifitime per day now, so I don't want to introduce breaking changes on the version 3.x releases.

gwbres:
Hello Chris,
it picked up a few lines that are not covered by a test case, i'll submit that tomorrow i'm heading to bed right now :) 
see you

gwbres:
@ChristopherRabotin ,

I forgot to increment a loop in a testbench somewhere.  
These loops iterate from 0.. to N , where N is the highest integer number representing the "latest GNSS" system known by this library. 
This is can of easy to forget since we don't have a macro to figure this number out easily (previously 7, now incremented to 8).

But that triggers `src/asn1der::test_encdec` correctly now which says a QZSS timescale was not preserved through the test.  
Obviously, the problem comes from what happens in `Epoch::from_der[u8 buffer]` but I have no idea what function this inherited feature relies on.  
Currently, all Integer or Str interpretation are properly coded IMO

src/timescale.rs
```diff
@@ -64,6 +64,7 @@ pub const UNIX_REF_EPOCH: Epoch = Epoch::from_tai_duration(Duration {
 });
 
 /// Enum of the different time systems available
+#[non_exhaustive]
```
ChristopherRabotin, line 67: Great! https://github.com/nyx-space/hifitime/issues/210#issuecomment-1475026427

ChristopherRabotin (approved):

Issue #234, opened 2023-05-26
MonthName from u8 


Massou31:
the implementation of `MonthName` from u8 is incorrect.

https://github.com/nyx-space/hifitime/blob/master/src/month.rs#L73

If the input is 12 then `January` is returned instead of `December`

PR #235, opened 2023-05-26
fix december for MonthName from u8

Massou31:
fixes https://github.com/nyx-space/hifitime/issues/234


ChristopherRabotin:
Oh, yeah, great point! The code in the month source file is also the least covered because it seemed so trivial and yet, there was a bug! Thanks for fixing it.

Issue #237, opened 2023-05-26
Refactor `Epoch` to keep track of a duration from time scale reference


ChristopherRabotin:
In a previous version of Hifitime 3.x, the Epoch didn't store the time scale in which it was defined. For that reason, all of the epochs were converted to TAI on initialization.

Now that Epoch stores its timescale, I don't think that the initializers should convert to TAI every time: this adds complexity, and quite frankly, makes it hard to follow what conversions are happening. I believe that confusion is the source of https://github.com/nyx-space/hifitime/issues/209.

Switching to defining an epoch as a duration in a given time scale is both simpler code and easier to keep track of the conversions. The downside here is that modifiers (like Modified Julian Dates) will still require the conversion since these modifiers include both another reference epoch and a time scale (unless these modifiers become time scales in their own right?!).

What I'm proposing therefore is the following new definition:

```rust
pub struct Epoch {
    pub duration_since_ref: Duration,
    pub time_scale: TimeScale,
}
```

Another option would be to specify both a time scale and a reference epoch in the `Epoch` where the time scale is used to convert only the tick rate and reference epoch to align when that tick rate starts. That option seems significantly more complicated though, and most time scales tick at the same rate as TAI/UTC.

@gwbres  what do you think?

ljbade:
I do like this idea, as sometimes you want to store GPS or GAL etc timestamp in a type with an Epoch, and then be able to inspect what the original time scale was later on.

It would be easy to then convert always to something like GPS (or TAI) time scale when needing them all in same timescale for doing calculations.

ChristopherRabotin:
We started working on it a few months ago here: https://github.com/nyx-space/hifitime/pull/241 . I paused the work when I thought that I might have uncovered a bug in the TDB reference epoch... but I'm not sure whether that's the case since the validation tests pass in flying colors and just one of the tests I wrote doesn't (if I remember the status correctly).

I'll pick this up after I release version 0.1 of [anise](https://github.com/nyx-space/anise/).

PR #238, opened 2023-05-27
epoch::from_duration slight improvements

gwbres:
Hello @ChristopherRabotin, 

this is an inquiry that might get pruned.  
i'm trying to see if there are any possibilities to simplify some conversions and construction ops.    

Right now the only thing I achieved is a slightly tighter Epoch::from_duration and use it everywhere.

Inquiring #237 and possible traits will most certainly have more impact

### DIFF
--- a/src/epoch.rs
+++ b/src/epoch.rs
@@ -263,20 +263,75 @@ impl Epoch {
         None
     }
 
-    /// Makes a copy of self and sets the duration and time scale appropriately given the new duration
+    /// Creates an epoch from given duration expressed in given timescale.
+    /// In case of ET, TDB Timescales, a duration since J2000 is expected.
     #[must_use]
-    pub fn from_duration(new_duration: Duration, time_scale: TimeScale) -> Self {
-        match time_scale {
-            TimeScale::TAI => Self::from_tai_duration(new_duration),
-            TimeScale::TT => Self::from_tt_duration(new_duration),
-            TimeScale::ET => Self::from_et_duration(new_duration),
-            TimeScale::TDB => Self::from_tdb_duration(new_duration),
-            TimeScale::UTC => Self::from_utc_duration(new_duration),
-            TimeScale::GPST => Self::from_gpst_duration(new_duration),
-            TimeScale::QZSST => Self::from_qzsst_duration(new_duration),
-            TimeScale::GST => Self::from_gst_duration(new_duration),
-            TimeScale::BDT => Self::from_bdt_duration(new_duration),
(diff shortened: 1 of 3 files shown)


src/epoch.rs
```diff
@@ -263,22 +263,77 @@ impl Epoch {
...
-            TimeScale::TDB => Self::from_tdb_duration(new_duration),
-            TimeScale::UTC => Self::from_utc_duration(new_duration),
-            TimeScale::GPST => Self::from_gpst_duration(new_duration),
-            TimeScale::QZSST => Self::from_qzsst_duration(new_duration),
-            TimeScale::GST => Self::from_gst_duration(new_duration),
-            TimeScale::BDT => Self::from_bdt_duration(new_duration),
+    /// Creates an epoch from given duration expressed in given timescale.
+    /// In case of ET, TDB Timescales, a duration since J2000 is expected.
+    #[must_use]
+    pub fn from_duration(new_duration: Duration, ts: TimeScale) -> Self {
```
ChristopherRabotin, line 269: If #237 architecture is adopted, this can simply become:

```rust
pub const fn from_duration(duration: Duration, ts: TimeScale) -> Self {
    Self {
        duration,
        time_scale: ts
    }
```

Then the conversion to/from different time scales only happens when another time scale is requested. For example, if someone did this:

```rust
let e = (366.days() + 1.hours()).ephemeris_time() // Create an ephemeris time epoch that's 2001-01-01T01:00:00 ET (because the ET reference is 01 Jan 2000)
let e_utc = e.into_time_scale(TimeScale::UTC) // This is when the computation between time scales happen
```

Note that I _think_ `into_...` works here given the [Rust naming convention](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv).

As I type this, two other things come to mind:
1. I think it would make sense for `Week` to be a `Unit` so one could do `5.weeks() + 6.hours()`: a week is always exactly 7 days, and I think we didn't add `Week` to `Unit` because it would have been a breaking change
2. The reference epoch should be defined for all TimeScales. The issue still comes with modifiers since there may be some use cases of `365.days().mjd_ephemeris_time()` where the reference epoch is Modified Julian Days... but that adds complexity and rigidity because people can't easily define new reference epochs... Maybe something `365.days().ephemeris_time_from(REF_EPOCH)` where `ref_epoch` is also an epoch that's added to the ephemeris_time reference epoch ... I don't know what's best!

gwbres:
moved to dev-gh-237 since i'm inquiring this topic specifically

PR #241, opened 2023-05-28
Refactor Epoch to its own time scale

gwbres:
Trying to see how complex answering to #237 would be.

* Renamed `Epoch::j1900_duration`
* Introduced/Renamed `Epoch::to_time_scale` : follow naming conventions
* `Epoch::with_time_scale` is just a construction helper (follows naming conventions too)
* (+) and (-) ops first invoke a `rhs.to_time_scale(self.ts)` so the computation is feasible
* We need common ground  for  `cmp`  and `eq` ops, but i'm not sure it really makes sense to go this far. 
Maybe strict equality/comparison between both self.duration and self.ts is enough. Right now, the current code produces a funny behavior where .eq() fails for two identical datetimes:

```
  left: `2020-01-01T00:00:37 UTC`,
 right: `2020-01-01T00:00:37 UTC`test asn1der::test_encdec ... ', src/epoch.rs
FAILED: 3241:5
```
* `Epoch::to_time_scale(TimeScale::TDB)`  and other time scales that need to refer to J2000 may not be feasible if we come from a datetime that is not past J2000 (i marked a `TODO`) 
* The ton of test failures comes from
  - `Epoch::maybe_from_gregorian` possibly broken, although I kept the previous logic and it should be closed to working
  - `Epoch::compute_gregorian`, used all over the place, might no longer work because self is no longer referenced to J1900


gwbres:
Ok i fixed several issues by auto deriving `Eq` and `Hash`.

NB: it is considered bad practice to implement yourself `Hash` if you don't implement yourself `Eq` or vice-versa.  
Unfortunately I cant' recall the reason for that.  
Also, current lib hashes the inner duration, but not the inner time scale: this will probably lead to serdes issues ?
 
I moved to auto derivation for Eq and Hash.

As far as PartialEq and Cmp ops go, we still implement because we need to convert to identical time scale, for a perfect duration comparison

src/epoch.rs
```diff
@@ -263,27 +272,109 @@ impl Epoch {
...
         }
     }
 
+    pub fn to_duration_since_j1900(&self) -> Duration {
+        self.to_time_scale(TimeScale::TAI).duration
+    }
+
+    #[must_use]
+    /// Converts self to another time scale
+    pub fn to_time_scale(&self, ts: TimeScale) -> Self {
```
ChristopherRabotin, line 291:
```suggestion
    pub fn into_time_scale(&self, ts: TimeScale) -> Self {
```

I think that `into...` is a better name

src/epoch.rs
```diff
@@ -263,27 +272,109 @@ impl Epoch {
...
+        match ts {
+            TimeScale::TAI => {
+                let mut new_epoch = self.clone();
+                // if previous time scale supported leap seconds: remove them
+                if self.time_scale.uses_leap_seconds() {
+                    new_epoch -= new_epoch.leap_seconds(true).unwrap_or(0.0) * Unit::Second;
+                }
+                // conversion to TAI: remove time scale reference point
+                new_epoch.duration -= ts.tai_reference_epoch().duration;
+                new_epoch.with_time_scale(TimeScale::TAI)
```
ChristopherRabotin, line 301:
```suggestion
                new_epoch.time_scale = TimeScale::TAI;
                new_epoch
```

src/epoch.rs
```diff
@@ -263,27 +272,109 @@ impl Epoch {
...
+                let duration_since_j2000 = tai_epoch - J2000_REF_EPOCH;
+                let seconds_since_j2000 = duration_since_j2000.to_seconds();
+                let gamma = Self::inner_g(seconds_since_j2000);
+                let delta_tdb_tai = gamma * Unit::Second + TT_OFFSET_MS * Unit::Millisecond;
+                tai_epoch += delta_tdb_tai;
+                tai_epoch += J2000_TO_J1900_DURATION; // TDB time scale is expressed past J2000
+                tai_epoch.with_time_scale(TimeScale::TDB)
+            }
+            ts => {
+                // first convert back to TAI
```
ChristopherRabotin, line 343: Yes, I think that's correct. If I understand this correctly, it means that only ET and TDB have a different "tick" rate then the other time scales, i.e. one second in {TAI, UTC, GPS, GST, TT, ...} is one second in all of the other time scales, apart for TDB and ET. I think that is correct.

src/deprecated.rs
```diff
@@ -78,14 +78,16 @@ impl Epoch {
...
 
-    #[must_use]
-    #[deprecated(
-        note = "Prefix for this function is now `to_` instead of `as_`.",
-        since = "3.5.0"
-    )]
-    pub fn as_duration_since_j1900_in_time_scale(&self, time_scale: TimeScale) -> Duration {
-        self.to_duration_since_j1900_in_time_scale(time_scale)
-    }
+    /*
```
ChristopherRabotin, line 81: I think that we can remove all of the `deprecated.rs` file since it'll be a new major release, we can clean that up.

gwbres:
OK so most errors come from faulty From_gregorian and Compute_to_gregorian functions.
Could you help me out on those ?

I think I'm on the right track for Self::from_gregorian to which we now must add a notion of time scale.
I test it with `cargo test gpst` if you read the first few lines of tests::epoch.rs:gpst, we have a simple use case.

In V3, everything is referenced to TAI_J1900. So we have += 1900 hardcoded at some point, in from_gregorian().  
I managed to get the timescale reference offset correctly. 
With this, for example, i'm able to have `cargo test test_serdes` to work. 
But for some reason, the equality of this simple 1 second into GPST test is failling. It's probably due to the equal operator implementation

ChristopherRabotin:
Sorry I dropped the ball on this, I totally forgot you were working on it. Let me see what I can do.

ChristopherRabotin:
The tests are _nearly_ there, but I think that the one test that fails actually shows an error nested deep in the code, and I can't seem to figure out how to fix it yet.

The conversion from a J1900 referenced epoch to a J2000 epoch is wrong by twelve hours. It so happens that the difference between these two references is indeed twelve hours: J1900 starts a noon but J2000 starts at midnight. 

What I don't understand is that the computation to/from ET/TDB (which are J2000 based) is correct according to all of the validation test cases. But when we compute the Gregorian time, the twelve hour difference is visible. So I don't know what is going on.

gwbres:
> The tests are _nearly_ there, but I think that the one test that fails actually shows an error nested deep in the code

Thank you for your time, I'll take a look into this next week end, see you

ChristopherRabotin:
@gwbres I might be able to pick up this work in September. Have you had time to review the issue I found ? I really hope that it's me misunderstanding something because the code is _very_ close to being complete if the time difference is not a bug.

gwbres:
Trying to get back to this, will keep you updated.

An hypothetical Hifitime V4 would help close [this issue](https://github.com/georust/rinex/issues/81) for example.

.gitignore
```diff
@@ -2,4 +2,7 @@
 **/*.rs.bk
 Cargo.lock
 .env
-dist/
\ No newline at end of file
+dist/
+.venv
+src/epoch.rs
```
gwbres, line 7: are we sure about this ?

src/epoch.rs
```diff
@@ -1146,19 +1243,18 @@ impl Epoch {
         };
 
         let (mut year, mut days_in_year) = div_rem_f64(days_f64, DAYS_PER_YEAR_NLD);
-        // TAI is defined at 1900, so a negative time is before 1900 and positive is after 1900.
-        year += 1900;
+        year += ts.ref_year(); // NB: this assumes all known time scales started after J1900
```
gwbres, line 1246: this is something to keep in mind

src/epoch.rs
```diff
@@ -1146,19 +1269,18 @@ impl Epoch {
         };
 
         let (mut year, mut days_in_year) = div_rem_f64(days_f64, DAYS_PER_YEAR_NLD);
-        // TAI is defined at 1900, so a negative time is before 1900 and positive is after 1900.
-        year += 1900;
+        year += ts.initial_year(); // NB: this assumes time scales all started after J1900
```
gwbres, line 1272: something to keep in mind

src/timescale/mod.rs
```diff
@@ -217,61 +192,45 @@ impl From<u8> for TimeScale {
...
-    for ts_u8 in 0..u8::MAX {
-        let ts = TimeScale::from(ts_u8);
-        let ts_u8_back: u8 = ts.into();
-        // If the u8 is greater than 5, it isn't valid and necessarily encoded as TAI.
-        if ts_u8 < 9 {
-            assert_eq!(ts_u8_back, ts_u8, "got {ts_u8_back} want {ts_u8}");
-        } else {
-            assert_eq!(ts, TimeScale::TAI);
+    #[test]
+    fn test_initial_year() {
```
gwbres, line 224: I refactored `ref_year()`, renamed to `initial_year()` which is intended to be used in Compute_from_gregorian, mostly when formatting.

src/epoch.rs
```diff
@@ -215,29 +256,40 @@ impl AddAssign<Duration> for Epoch {
...
-    }
-}
-
-impl Hash for Epoch {
-    fn hash<H: Hasher>(&self, hasher: &mut H) {
-        self.duration_since_j1900_tai.hash(hasher);
+        if self.time_scale == other.time_scale {
+            self.duration == other.duration
+        } else {
+            self.duration == other.to_time_scale(self.time_scale).duration
```
gwbres, line 262: why not use this ?

gwbres:
@ChristopherRabotin,

I think I got the Gregorian methods to work, thanks to the new `TimeScale::initial_year()` helper.  
Currently I disabled TT, ET and similar timescales that are more complex and which I'm not familiar with.
I'm trying to get UTC, TAI, GNSS to work (simpler). 

If you run the following command in the current state, I don't understand why this test no longer passes : 

```bash
cargo test -- all-features -- --no-capture utc_tai
```

src/timescale/mod.rs
```diff
@@ -126,49 +117,34 @@ impl TimeScale {
...
+            _ => (1900, 01, 01, 00, 00, 00, 00),
         }
     }
-}
 
-impl fmt::LowerHex for TimeScale {
-    /// Prints given TimeScale in RINEX format
-    /// ie., standard GNSS constellation name is preferred when possible
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+    pub(crate) const fn ref_hour(&self) -> i64 {
```
gwbres, line 144: can probably get rid of ref_hour() now that we have .decompose()

src/timescale/mod.rs
```diff
@@ -126,49 +117,34 @@ impl TimeScale {
...
-            Self::TDB => write!(f, "TDB"),
-            Self::UTC => write!(f, "UTC"),
-            Self::GPST => write!(f, "GPST"),
-            Self::GST => write!(f, "GST"),
-            Self::BDT => write!(f, "BDT"),
-            Self::QZSST => write!(f, "QZSST"),
+            Self::GPST => (1980, 01, 06, 00, 00, 00, 00),
+            Self::BDT => (2005, 01, 01, 00, 00, 00, 00),
+            Self::GST => (1999, 21, 07, 00, 00, 00, 00),
+            _ => (1900, 01, 01, 00, 00, 00, 00),
```
gwbres, line 140: hard to create this yet "simple" requirement...

src/epoch.rs
```diff
@@ -1815,68 +1847,49 @@ impl Epoch {
...
     }
 
     #[must_use]
     /// Returns this time in a Duration past J1900 counted in TAI
-    pub const fn to_tai_duration(&self) -> Duration {
-        self.duration_since_j1900_tai
+    pub fn to_tai_duration(&self) -> Duration {
+        let mut dur = self.duration + self.time_scale.tai_reference_epoch().duration;
+        if self.time_scale.uses_leap_seconds() {
+            dur += self.leap_seconds(true).unwrap_or(0.0) * Unit::Second;
```
gwbres, line 1878: Leap second compensation is clearly needed, If i do it manually, it works.
But the `leap_seconds(true)` is causing a stack overflow, this has never cause problems before ?

gwbres: because _to_tai_duration_ is called inside  _leaps_seconds_with_

gwbres:
So Epoch.leap_seconds() needs to_tai_duration(), 
to_time_scale() needs leap_seconds() ....

ChristopherRabotin:
Thanks for all your help here, I'll run the tests locally soon. Are the UTC and GNSS time scales nearly working? I can handle the TT, ET, TDB.

gwbres:
> I don't understand why the library always prints UTC, it is very confusing.
> For example `epoch = Epoch::from_gpst_seconds(1.0)` prints 1980-01-06:01 UTC, but we want 1980-01-06:01 GPST, basically the gregorian date with emphasis of which time scale it is based on ?

This is due to the Display and Debug implementations. Although I would be OK with a diffrent debug behavior, I don't think the current Display behavior is right. What is your opinion on that ? it's confusing when dealing with a specific time scale, to see something expressed in UTC 

> Are the UTC and GNSS time scales nearly working? I can handle the TT, ET, TDB.

Hard to say. For me, the followings need to be solved :

- How an epoch is displayed
- To and From gregorian() have issues, which creates problems in places you would not expect. 
For me it's ok to have gregorian datetimes in timescales other than UTC
- PartialEq : are we ok on this ? 
- `to_time_scale()` and its relationship with `to_xxxx_duration()` and  `.leap_seconds()`.
For me, it is possible to have a  `to_time_scale(TAI)` which serves as the foundation to all other conversions.
It's somewhat what i'm trying to do

ChristopherRabotin:
I'm happy changing the behavior of `Display` to display the epoch in its own time scale. Maybe `Debug` could show the Duration in that time scale, e.g. `3455 days 2 h 15 ns since GPS epoch`. I think that provides useful debugging information.

For the `to_time_scale()`, converting via TAI is what we essentially have for version 3.x and it makes things complicated sometimes. I think that there was a commit a few months ago on this branch where I did a pretty good implementation of that function and I think it worked for most cases.

What's the issue with PartialEq ? I think that this makes sense, no?
```
//// If one of the two time scales does not include leap seconds,
//// we always convert the time scale with leap seconds into the
//// time scale that does NOT have leap seconds.
```

gwbres:
> I think that there was a commit a few months ago on this branch where I did a pretty good implementation of that function and I think it worked for most cases

Reverting to that state (now)

gwbres:
> What's the issue with PartialEq ? I think that this makes sense, no?

I just don't see why having a case specific to Leap Seconds, 
while we have a ` else {} `  branch that will manage it anyway :

```rust
 266             if self.time_scale.uses_leap_seconds() != other.time_scale.uses_leap_seconds() {
 267                 if self.time_scale.uses_leap_seconds() {
 268                     self.to_time_scale(other.time_scale).duration == other.duration
 269                 } else {
 270                     self.duration == other.to_time_scale(self.time_scale).duration
 271                 }
 272             } else {
 273                 // Otherwise it does not matter
                          // --> this can manage the leap seconds
 274                 self.duration == other.to_time_scale(self.time_scale).duration
 275             }
```

gwbres:
>  Maybe `Debug` could show the Duration in that time scale, e.g. `3455 days 2 h 15 ns since GPS epoch`. I think that provides useful debugging information.

I think that is a great idea

gwbres:
Can you help me fix Epoch::compute_gregorian and Epoch::maybe_from_gregorian ?

In Epoch::compute_gregorian(duration, TimeScale) : I introduce a time_scale.duration.decompose() to help recreate a gregorian date, we should be able to achieve that.

In Epoch::maybe_from_gregorian, I think it is much simpler than that. We just assume the user is defining the datetime he wants, inside that time scale, no need to offset anything, we just take the values as is and set the time scale as desired

gwbres:
@ChristopherRabotin 

I just pushed a commit that modifies the std::fmt::Debug behavior for Epoch.  
It now prints the duration in the associate time scale.

I think it's a very good idea you had here, it might make things so much easier when debugging calculations. 
std::fmt::Display is the one intended for the end user

gwbres:
Hello @ChristopherRabotin 

Can you please review what's being tested in tests/epoch.rs `fn test_timescale_recip` ?

image

There is something terribly wrong with this test, either how it's designed or how it's operated.

How can a duration ("from_dur") built from a different timescale, be the same as a UTC duration.
But I get what you're trying to do. You're trying to see if a reciprocal cast leads to a strictly identical duration.
But Epochs and Durations are two different things. We need to compare durations here, not Epochs.

I just pushed a commit that modifies the PartialEq behavior (see Epoch::PartialEq), and this test is the only one that used to pass, now fails due to that modification. Might be a good thing ?

Any chances to make progress on this topic in the near future?

ChristopherRabotin:
I think that this test is obsolete now with the new design: it used to be important because we would convert from the initialization time scale to TAI every time, so the test would make sure that if we initialized at Epoch X in time scale T, when it was converted to TAI and then requested back in T, then the same epoch X would be returned.

I think this test can be deleted now.

ChristopherRabotin:
I'm OK with the changes to PartialEq. The previous implementation was trying to be clever, but maybe the better use case is to ask the user to convert both epochs to the same time scale before comparing them.

ChristopherRabotin:
@gwbres I have a bit more clarity on my schedule, and I think I can pick up the work on this in February. I've also been thinking more about this, and I now think that this change can be (and should be) merged into the same 3.x version of hifitime. Although in theory any changes to the enums are considered breaking changes, I think it's unlikely that users perform a `match` on the time scale, and as such, the changes introduced in what we initially called `4.0.0` are only "potentially but unlikely" breaking changes. Keeping everything in version 3.x would mean that we have to keep the older functions available, and just mark them as deprecated.

What do you think?

gwbres:
It sounds good to me.
The only thing to be careful with this new definition, is the `Eq` operator. 
Auto derivation is not possible because auto derivation :

```rust
#[derive(PartialEq)]
pub struct Epoch {
    Duration,
    TimeScale
```

would return true when dur(a) == dur(b) && timescale(a) == timescale(b).
This behavior is very wrong because it is possible to represent the same "time" with different durations in different timescale.
I'm very used to working with GPST for example, let's take the Midnight UTC example, GPST is late by the amount of leap second, so midnight UTC and 23:59:23 GPST are the same instant.
Conclusion : Eq needs to be manually implemented and requires the conversion into a common timescale

ChristopherRabotin:
Good reminder!

gwbres:
I don't know if continuing this branch is the way to go, it is one possibility.
Once the redesign is stabilized, I can have my tools use a local version of this branch. We use the library extensively, so it will be another good indication

ChristopherRabotin:
This should not have been closed ! I didn't expect that delete the 4.0.0-dev branch would close this PR. Can you re-open it directly against `master`?

gwbres:
Sure, I will take care of this in the next few days


Guillaume W. Bres
Software engineer
***@***.***>

Issue #267, opened 2023-12-12
Support both %yy and %YYYY format ?


gwbres:
https://github.com/nyx-space/hifitime/blob/f0f7fd0fbb1f65d43592f656f97e6d6f1c4048e7/src/efmt/format.rs#L37

Would it be possible to introduce a %y format to describe year on two digits, which you sometimes encounter ?
in which case we would consider Y=0 = 2000. I presume that's what python datetime does

ChristopherRabotin:
Yes, we should have this. It's part of the C89 standard as well. I think it
should be uppercase and lowercase Y.

ChristopherRabotin:
With pleasure! I think that "all" that needs to be done is to add a parsing
token called YearShort (in the format module), and implement the parsing
and formatting. Just by adding that enum variant, that should cause a few
compilation errors, and those are where you'll need to implement the
parsing or formatting of the "short year." I'd prefer not changing the
other enum variant because that would be a breaking change for version 3.

And soon I'll need to pick up the work on version 4. I think your PR was
very close a few months ago, but I got confused about only one test...

gwbres:
I just opened a PR with a first test that does work

ChristopherRabotin:
Implemented in #268, thanks Guillaume !

PR #268, opened 2023-12-16
Introduce %y formatter

gwbres:
* Support %yy year on two digits

### DIFF
--- a/src/efmt/format.rs
+++ b/src/efmt/format.rs
@@ -35,6 +35,7 @@ const MAX_TOKENS: usize = 16;
 /// | Token | Explanation | Example | Notes
 /// | :-- | :-- | :-- | :-- |
 /// | `%Y` | Proleptic Gregorian year, zero-padded to 4 digits | `2022` | (1) |
+/// | `%y` | Proleptic Gregorian year after 2000, on two digits | `23` | N/A |
 /// | `%m` | Month number, zero-padded to 2 digits | `03` for March | N/A |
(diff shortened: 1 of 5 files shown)


ChristopherRabotin:
Looks good, thanks!

Closes #267

src/efmt/format.rs
```diff
@@ -228,6 +230,12 @@ impl Format {
                 let sub_str = &s[prev_idx..end_idx];
 
                 match prev_token {
+                    Token::YearShort => {
+                        decomposed[0] = sub_str
+                            .parse::<i32>()
+                            .map_err(|_| Errors::ParseError(ParsingErrors::ValueError))?
+                            + 2000;
```
gwbres, line 237: @ChristopherRabotin shouldn't `sub_str` be trimmed() here, so `.parse()` never fails ? 
In other terms, may whitespaces exist prior next separator

ChristopherRabotin: Ah, I didn't think of this. I forget how it works. Don't we parse token by token and therefore it isn't possible to have a space? Do the other cases have a sub str? I'm afraid I'm busy this weekend and can't look into this.

gwbres:
Hello, 
would it be possible to release this ?

ChristopherRabotin:
Sure! I'll release it today.

gwbres:
Hello Chris,
this won't buy me anything if the other issue is not solved, because my use case is to parse more easily GNSS timestamps and use the Day of Year aswell. So in my case, it can wait for its resolution

PR #271, opened 2023-12-30
docs(fix): fixed reversed description of duration magnitude

ethanmsl:
Just a fix of what I believe is a typo, but also a couple out-of-code comments re: possible changes.

Fix: the docs said that -1 century < (-1 century + 1 ns), which I believe is the opposite of what you meant.  (Had to download library and check though since, per the point of explaining, it could've gone either way :)

A couple other things that might be worth changing/clarifying:

1. > It was decided that the nanoseconds corresponds to the nanoseconds _into_ the current century.
    - I think that's *also* untrue, yes?  Nanoseconds are toward "the future" as it were.  Rather than toward any period (absolute or dynamic).  Technical correctness aside, I also found that phrasing confusing.  (But I'm not a space-related scientist, so maybe this is common phrasing in your field. ?)

2. > The negative durations can be mentally modeled "BC" years. One hours before 01 Jan 0000, it was "-1" years but  365 days and 23h into the current day.
    - This is helpful, but might it be simpler to just refer to the sign of both durations *and* components as pointing to the past (back) or future (forward)?  This also sets up the clean reading of +/- centuries, + nanoseconds as just being negative + positive.
    
Just figured I'd bring it up in case you thought minor re-wording could help, since I was making a (micro) PR.
(I've only just found your crate, so I have zero expertise here and may be misunderstanding its intended interpretation.)

### DIFF
--- a/src/duration.rs
+++ b/src/duration.rs
@@ -54,7 +54,7 @@ pub const NANOSECONDS_PER_CENTURY: u64 = DAYS_PER_CENTURY_U64 * NANOSECONDS_PER_
 /// **Important conventions:**
 /// 1. The negative durations can be mentally modeled "BC" years. One hours before 01 Jan 0000, it was "-1" years but  365 days and 23h into the current day.
 /// It was decided that the nanoseconds corresponds to the nanoseconds _into_ the current century. In other words,
-/// a duration with centuries = -1 and nanoseconds = 0 is _a smaller duration_ (further from zero) than centuries = -1 and nanoseconds = 1.
+/// a duration with centuries = -1 and nanoseconds = 0 is _a greater duration_ (further from zero) than centuries = -1 and nanoseconds = 1.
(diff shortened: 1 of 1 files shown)


ChristopherRabotin:
Hi Ethan,

First off, thanks for the sponsorship!

Secondly, thank you for the fix. Yes, that was a typo. Here it is in action in Python:
```
In [10]: (Unit.Century*-1 + Unit.Nanosecond*1).to_parts()
Out[10]: (-1, 1)

In [11]: (Unit.Century*-1).to_parts()
Out[11]: (-1, 0)

In [12]: Unit.Century*-1 < Unit.Century*-1 + Unit.Nanosecond*1
Out[12]: True
```

I fully agree with you that the documentation is confusing: I recall being confused when I was proof-reading it that part of it. At the same time, I couldn't quite figure out how to write it differently. With more perspective now, maybe can I find another way to explain how it works.

PR #279, opened 2024-01-15
Draft: replace std::time:SystemTime with web_time::SystemTime

### LINKED ISSUE #278: Epoch::now() is returning faulty time (wasm/web-sys)
#262 Adds wasm support. But the duration obtained through web-sys is currently wrong.
It uses Performance.now (https://developer.mozilla.org/en-US/docs/Web/API/Performance/now) which returns the milliseconds since the page/worker has loaded.

See the current implementation in https://github.com/nyx-space/hifitime/pull/262/files#diff-2effce6c18e7a6bda6cd5148cecf1af4951be035c6c03b6665e0275539c7f5eeR8-R12 
- we always start at time since Unix Epoch = 0 (time since page navigation/worker load)
- we multiply milliseconds * Unit::Second which results in 1000 seconds being added every second
- we lose resolution (reduced to seconds)

Instead we should use something like https://lib.rs/crates/web-time, or adopt it's implementation for this use-case
> This implementation uses [Performance.now()](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now) for [Instant](https://doc.rust-lang.org/std/time/struct.Instant.html) and [Date.now()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now) for [SystemTime](https://doc.rust-lang.org/std/time/struct.SystemTime.html) to offer a drop-in replacement that works in browsers

karpfediem:
Would fix #278 
Allows consistent usage of SystemTime between targets (including WASM).

I replaced all uses of std::time::SystemTime with web_time::SystemTime, and removed the overloaded `duration_since_unix_epoch` method for the WASM target.

I haven't vetted the web_time library for other uses, but it seems to be working fine. 
So this is just a proposal and might need more discussion/research before including as dependency.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ lexical-core = { version = "0.8.5", default-features = false, features = [
 reqwest = { version = "0.11", features = ["blocking", "json"], optional = true }
 tabled = { version = "0.15.0", optional = true }
 openssl = { version = "0.10", features = ["vendored"], optional = true }
+web-time = { version = "1.0.0" }
 
 [target.wasm32-unknown-unknown.dependencies]
 js-sys = { version = "0.3" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ lexical-core = { version = "0.8.5", default-features = false, features = [
 reqwest = { version = "0.11", features = ["blocking", "json"], optional = true }
 tabled = { version = "0.15.0", optional = true }
 openssl = { version = "0.10", features = ["vendored"], optional = true }
-web-time = { version = "1.0.0" }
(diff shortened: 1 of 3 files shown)


ChristopherRabotin:
Thanks for the pull request that rapidly address the bug you found.

I think that the crate needs to be marked as optional and enabled only when the std feature is enabled. Then, I think the tests will pass.

karpfediem:
> Thanks for the pull request that rapidly address the bug you found.
> 
> I think that the crate needs to be marked as optional and enabled only when the std feature is enabled. Then, I think the tests will pass.

Oh i see now, thanks for the heads-up.
And it's a simple fix, happy to help :)

PR #280, opened 2024-02-08
Refactor epoch to keep time in its own time scale

ChristopherRabotin:
Reopening https://github.com/nyx-space/hifitime/pull/241


ChristopherRabotin:
Hi @gwbres I have some design questions for you.

First, importantly, I think that we should merge this as part of version 4.0 and I'll take this opportunity to clean up the code from the what is deprecated, refactor the modules to make them more maintainable (it isn't OK to have 4000 lines in a single module, I just added things continuously and never cleaned this up), and general chores.

Secondly, here are some questions:
1. In the new implementation proposed here (which I like a lot), the `to_time_scale` function currently only performs the conversion between time scales _if_ the original time scale and the new one are different. I think it makes sense that we should avoid computations if we can. But shouldn't we expect the behavior of the code to be the same whether or not we perform this computation? My first thought says "yes" but I'm having second thoughts so I'm interested in what you think. For example, in the [`ts_over_leap_second`](https://github.com/nyx-space/hifitime/blob/822c0a0ee4da21f385bc520547431fd1b3ed7d6f/src/timeseries.rs#L382) time series test, the result is different if I add `&& false` to line 339 of epoch.rs or not.
2. I'll take advantage of this PR to also fix #282 raised in the discussions. Any objection to that?

That's all for now. But this time, I really want to pick up this work because the list of issues open keeps growing and I want to resolve them.

ChristopherRabotin:
Moved to #289

Issue #244, opened 2023-06-17
Bug Report: Identified issues with arithmetic overflow, unreachable code, unwrap errors, and array out-of-bounds and so on


xizheyin:
## Description
HI! I am using my fuzz testing tool to test this library, and so far I have found 15 different bugs. Among them, there are 5 array out-of-bounds errors, 3 string encoding errors, 1 unwrap error, 1 unreachable code bug, and 5 arithmetic overflow bugs. Below is the list of errors. Please review them and check if any modifications are needed. The replay files are all stored in [this repository](https://github.com/XizheYin-NJU/replay_files_hifitime).

### 1. Array out-of-bounds error
error message:
```
thread 'main' panicked at 'begin <= end (21 <= 20) when slicing `94-11-05T08:15:34.0-:0`', /home/yxz/.cargo/registry/src/mirrors.ustc.edu.cn-61ef6e0cd06fb9b8/hifitime-3.8.2/src/epoch.rs:993:43
```
source code:
![image](https://github.com/nyx-space/hifitime/assets/62123683/4ae47f3e-cab2-4aba-8615-9359a8eb1030)

### 2. Array out-of-bounds error
error message:
```
thread 'main' panicked at 'index out of bounds: the len is 16 but the index is 16', /home/yxz/.cargo/registry/src/mirrors.ustc.edu.cn-61ef6e0cd06fb9b8/hifitime-3.8.2/src/efmt/format.rs:400:25
```
source code:
![image](https://github.com/nyx-space/hifitime/assets/62123683/03aebae3-6b52-4586-b52e-1ade465c9f34)

### 3. Array out-of-bounds error
```
thread 'main' panicked at 'index out of bounds: the len is 16 but the index is 16', /home/yxz/.cargo/registry/src/mirrors.ustc.edu.cn-61ef6e0cd06fb9b8/hifitime-3.8.2/src/efmt/format.rs:488:25
```
source code:
![image](https://github.com/nyx-space/hifitime/assets/62123683/2a6a104e-de5c-4bea-9f6c-ebf9da054e43)

### 4. Array out-of-bounds error
error message:
```
thread 'main' panicked at 'index out of bounds: the len is 16 but the index is 16', /home/yxz/.cargo/registry/src/mirrors.ustc.edu.cn-61ef6e0cd06fb9b8/hifitime-3.8.2/src/efmt/format.rs:424:25
```
source code：
[...]

PR #283, opened 2024-03-12
Fix panic caused by arithmetic overflow

cardigan1008:
Fix 11. and 12. Arithmetic overflow bug and  in #244 
Consider using saturating_sub and saturating_mul to avoid the panic.

The problem lies in https://github.com/nyx-space/hifitime/blob/822c0a0ee4da21f385bc520547431fd1b3ed7d6f/src/epoch.rs#L735
where `year_since_1900` is overflowing to i32.

https://github.com/nyx-space/hifitime/blob/822c0a0ee4da21f385bc520547431fd1b3ed7d6f/src/epoch.rs#L736
where `365 * yeaer_since_1900` is overflowing to i32.

I've opted to use `saturating_sub` and `saturating_mul`. These functions ensure that there's a maximum boundary, preventing overflow panic.

PS: As a newcomer to the Rust project, I apologize for any inconvenience this may cause.

### DIFF
--- a/src/epoch.rs
+++ b/src/epoch.rs
@@ -737,7 +737,7 @@ impl Epoch {
             Some(years_since_1900) => match years_since_1900.checked_mul(365) {
                 None => return Err(Errors::Overflow),
                 Some(days) => (years_since_1900, Unit::Day * i64::from(days)),
-            }
+            },
         };
 
         // count leap years
--- a/src/epoch.rs
+++ b/src/epoch.rs
@@ -732,8 +732,8 @@ impl Epoch {
             return Err(Errors::Carry);
         }
 
-        let years_since_1900 = year - 1900;
-        let mut duration_wrt_1900 = Unit::Day * i64::from(365 * years_since_1900);
+        let years_since_1900 = year.saturating_sub(1900);
+        let mut duration_wrt_1900 = Unit::Day * i64::from(years_since_1900.saturating_mul(365));
 
         // count leap years
         if years_since_1900 > 0 {
--- a/src/epoch.rs
+++ b/src/epoch.rs
@@ -732,8 +732,13 @@ impl Epoch {
             return Err(Errors::Carry);
         }
 
-        let years_since_1900 = year.saturating_sub(1900);
-        let mut duration_wrt_1900 = Unit::Day * i64::from(years_since_1900.saturating_mul(365));
+        let (years_since_1900, mut duration_wrt_1900) = match year.checked_sub(1900) {
+            None => return Err(Errors::Overflow),
+            Some(years_since_1900) => match years_since_1900.checked_mul(365) {
+                None => return Err(Errors::Overflow),
+                Some(days) => (years_since_1900, Unit::Day * i64::from(days)),
+            }
+        };
 
         // count leap years
         if years_since_1900 > 0 {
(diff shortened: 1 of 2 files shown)


gwbres (approved):

ChristopherRabotin (dismissed):

ChristopherRabotin:
I'm proposing another way to fix this overflow bug because hifitime should error not hide these errors. Instead, could you apply the following patch to your branch (I wasn't sure if I had push permissions to your branch).

This patch reports an error to prevent an underflow or overflow, instead of hiding the error.

```diff
--- a/src/epoch.rs
+++ b/src/epoch.rs
@@ -732,8 +732,13 @@ impl Epoch {
             return Err(Errors::Carry);
         }
 
-        let years_since_1900 = year.saturating_sub(1900);
-        let mut duration_wrt_1900 = Unit::Day * i64::from(years_since_1900.saturating_mul(365));
+        let (years_since_1900, mut duration_wrt_1900) = match year.checked_sub(1900) {
+            None => return Err(Errors::Overflow),
+            Some(years_since_1900) => match years_since_1900.checked_mul(365) {
+                None => return Err(Errors::Overflow),
+                Some(days) => (years_since_1900, Unit::Day * i64::from(days)),
+            },
+        };
 
         // count leap years
         if years_since_1900 > 0 {
diff --git a/src/errors.rs b/src/errors.rs
index 473b05d..b8d9e03 100644
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -76,10 +76,7 @@ impl fmt::Display for Errors {
             Self::ConversionOverlapError(hi, lo) => {
                 write!(f, "hi and lo values overlap: {}, {}", hi, lo)
             }
-            Self::Overflow => write!(
-                f,
-                "overflow occurred when trying to convert Duration information"
-            ),
+            Self::Overflow => write!(f, "prevented overflow/underflow"),
             Self::SystemTimeError => write!(f, "std::time::SystemTime returned an error"),
         }
     }
```

cardigan1008:
Thank you for your patch, @ChristopherRabotin ! I've been actively working on automatically fixing Rust panics, and this patch is a result of my efforts using a simple tool. I came across #244 and tried to apply the tool on these panics.
I'm curious about whether the original patch is suitable for addressing a panic, especially considering that I've come across other patches that utilize saturating_sub/mul. Your response would be immensely helpful for me to further my research! However, I understand if this discussion might be disruptive, and I hope it doesn't inconvenience you.

ChristopherRabotin:
Hi Yunbo,

First off, thanks for contributing, I'm always happy to review and see how people use hifitime and they want to improve on it (even if I'm not always the fastest reviewer as gwbres knows).

In general, I think that the checked_ call works for most applications. That's why I had initially approved the changes before thinking about it some more during the day. In the case of hifitime however, and especially in this case where we try to convert a potentially wrong date, we want to fail early and give an indication of what has happened, and prevent the user from thinking that the date they provided was correct when in fact the library hid the error and uses the minimum valid date for example.

I hope this helps, and again, thanks for your contribution!

cardigan1008:
Thank you for taking the time to review, and I appreciate your insights. I completely understand your perspective, especially regarding the importance of early failure detection. It's crucial to provide clear feedback to users and prevent any confusion or incorrect assumptions about the validity of their input.

Your feedback is invaluable, and I'll make sure to keep it in mind for future research. Once again, thank you for your support and guidance!

cardigan1008:
> In general, I think that the checked_ call works for most applications. That's why I had initially approved the changes before thinking about it some more during the day.

By the way, I noticed your comment here about the checked_ call. Just to clarify, did you mean to refer to the saturating_ call instead?

PR #285, opened 2024-03-19
Fix panic caused by unwrap

cardigan1008:
Fix 8.Unwrap error in #244 

In the replay file, `from_format_str` takes "%" as format_str, thus causing the format returned by `from_str` empty.

https://github.com/nyx-space/hifitime/blob/822c0a0ee4da21f385bc520547431fd1b3ed7d6f/src/efmt/format.rs#L159

This results in the items in `parse` being empty, which causes an unwrap error. In my opinion, it should be classified as an UnknownFormat error since it's related to format.

### DIFF
--- a/src/efmt/format.rs
+++ b/src/efmt/format.rs
@@ -156,7 +156,10 @@ impl Format {
         // Previous index of interest in the string
         let mut prev_idx = 0;
         let mut cur_item_idx = 0;
-        let mut cur_item = self.items[cur_item_idx].unwrap();
+        let mut cur_item =  match self.items[cur_item_idx] {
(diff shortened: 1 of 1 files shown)


ChristopherRabotin:
Thanks for your contributions, Yunbo, I finally fixed the CI pipelines, so I've merged all of your outstanding PRs.

cardigan1008:
> Thanks for your contributions, Yunbo, I finally fixed the CI pipelines, so I've merged all of your outstanding PRs.

Thanks for your reviews too! They're a lot of help to the panic repair tool I've been working on.

PR #289, opened 2024-03-29
4.0.0 dev gh 237

ChristopherRabotin:
Reopening the work on #237 . 

@gwbres I was not able to push to your fork of the repo anymore, so I've pushed a new branch. I think you have write access to hifitime, so don't hesitate to work from here.

https://github.com/nyx-space/hifitime/pull/280#issuecomment-2026613195


src/epoch.rs
```diff
@@ -3266,12 +3266,21 @@ fn div_rem_f64_test() {
 }
 
 #[test]
-fn test_days_tdb_j2000() {
+fn test_days_et_j2000() {
```
ChristopherRabotin, line 3306: I'm concerned about this test... why did it work before?!

tests/epoch.rs
```diff
@@ -889,11 +907,34 @@ fn test_from_str() {
 }
 
 #[test]
-fn test_from_str_tdb() {
-    use core::str::FromStr;
-
-    let greg = "2020-01-31T00:00:00 TDB";
-    assert_eq!(greg, format!("{:e}", Epoch::from_str(greg).unwrap()));
+fn test_timescale_leapsec() {
```
ChristopherRabotin, line 910: @gwbres I don't understand this test that you added. Why should the duration be the number of leap second difference? In the `Sub` impl, we convert to the other time scale: https://github.com/nyx-space/hifitime/blob/5ac34a21180163897db6070f9c9d414bb321a845/src/epoch.rs#L166 . So shouldn't this always be zero?

ChristopherRabotin:
@gwbres , I'll be merging this once the tests go through if that's OK. The idea is to get this merged sooner since you've started this work so long ago, and then I have a lot more refactoring I want to make, in addition to fixing a bunch of the bugs that have been reported in the past year. Let me know if that's OK for you.

ChristopherRabotin:
Follow-on tasks: #294 , #295.

ChristopherRabotin:
Closes #237 .

Issue #261, opened 2023-11-27
jde_utc_days Error 


adcl-0:
if I create Epoch for dates of 2023 the represantation of date when printed is okay and the jde_utc is also correct.
but when I do that same for dates of year 1608 the representation of dates is wrong but jde calculation is still correct.
but as soon as i try the same for years 1607 and before both the jde calculation as well as the date representation is wrong.
is it a bug or something I am not able to understand.

I haven't tried the python library I just tried the rust 
Thanks 
Following is an example code and the result :
```rust
use hifitime;

pub fn jde_test() {
    let mut test: hifitime::Epoch;
    println!("Dates and JDE for year 1607");
    for i in 25..=31 {
        test = hifitime::Epoch::from_gregorian_utc(1607, 12, i, 0, 0, 0, 0);
        println!("{} - {}", test, test.to_jde_utc_days());
    }
    println!("Dates and JDE for year 1608");
    for i in 1..=5 {
        test = hifitime::Epoch::from_gregorian_utc(1608, 1, i, 0, 0, 0, 0);
        println!("{} - {}", test, test.to_jde_utc_days());
    }
    println!("Dates and JDE for year 2023");
    for i in 1..=5 {
        test = hifitime::Epoch::from_gregorian_utc(2023, 1, i, 0, 0, 0, 0);
        println!("{} - {}", test, test.to_jde_utc_days());
    }
}
```
```shell
Dates and JDE for year 1607
-3277145-01-01T00:00:00 UTC - -1194435819.5
-3277145-01-01T00:00:00 UTC - -1194435818.5
-3277145-01-01T00:00:00 UTC - -1194435817.5
-3277145-01-01T00:00:00 UTC - -1194435816.5
-3277145-01-01T00:00:00 UTC - -1194435815.5
-3277145-01-01T00:00:00 UTC - -1194435814.5
-3277145-01-01T00:00:00 UTC - -1194435813.5
Dates and JDE for year 1608
1592-01-01T00:00:00 UTC - 2308369.5
1592-01-01T00:00:00 UTC - 2308370.5
1592-01-01T00:00:00 UTC - 2308371.5
1592-01-01T00:00:00 UTC - 2308372.5
1592-01-01T00:00:00 UTC - 2308373.5
Dates and JDE for year 2023
2023-01-01T00:00:00 UTC - 2459945.5
2023-01-02T00:00:00 UTC - 2459946.5
2023-01-03T00:00:00 UTC - 2459947.5
[...]

PR #303, opened 2024-05-19
DRAFT -- Fix conversion to Gregorian

ChristopherRabotin:
Issue #261 made it clear that there were bugs in the conversion from a duration in a given time scale into its Gregorian representation. This PR tries to solve this by rewriting that algorithm and extending the tests to include many reciprocal tests.

As of now, this branch _does_ not solve all of the issues (hence the draft status), but I'm seeking a review because I can't seem to understand why the algorithm does not always work: from what I can tell, it is the inverse algorithm from the one that converts a Gregorian date to a duration so I expect it to work...

@gwbres , could you have a look at this proposed implementation to see if you agree with the new implementation?

Thanks

### DIFF
--- a/src/duration/mod.rs
+++ b/src/duration/mod.rs
@@ -692,7 +692,15 @@ impl fmt::Display for Duration {
             }
 
             let values = [days, hours, minutes, seconds, milli, us, nano];
-            let units = ["days", "h", "min", "s", "ms", "μs", "ns"];
+            let units = [
+                if days > 1 { "days" } else { "day" },
+                "h",
+                "min",
+                "s",
+                "ms",
+                "μs",
+                "ns",
+            ];
 
             let mut insert_space = false;
             for (val, unit) in values.iter().zip(units.iter()) {
@@ -767,7 +775,7 @@ mod ut_duration {
     fn test_serdes() {
         for (dt, content) in [
             (Duration::from_seconds(10.1), r#""10 s 100 ms""#),
-            (1.0_f64.days() + 99.nanoseconds(), r#""1 days 99 ns""#),
+            (1.0_f64.days() + 99.nanoseconds(), r#""1 day 99 ns""#),
             (
                 1.0_f64.centuries() + 99.seconds(),
                 r#""36525 days 1 min 39 s""#,
--- a/src/epoch/gregorian.rs
+++ b/src/epoch/gregorian.rs
@@ -61,18 +61,20 @@ impl Epoch {
             }
         }
 
-        if is_leap_year(year) && days_in_year > CUMULATIVE_DAYS_FOR_MONTH[2] as f64 {
-            days_in_year -= 1.0;
-        }
+        let cumul_days = if is_leap_year(year) {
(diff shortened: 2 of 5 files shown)


src/epoch/gregorian.rs
```diff
@@ -269,45 +251,43 @@ impl Epoch {
...
-            Some(years_since_ref) => match years_since_ref.checked_mul(365) {
+            Some(years_since_ref) => match years_since_ref.checked_mul(DAYS_PER_YEAR_NLD as i32) {
                 None => {
                     return Err(EpochError::Duration {
                         source: DurationError::Overflow,
                     })
                 }
                 Some(days) => Unit::Day * i64::from(days),
             },
-        } - time_scale.gregorian_epoch_offset();
```
ChristopherRabotin, line 280: Moved to the end of the function to make it easier to write the reverse algorithm.

src/epoch/gregorian.rs
```diff
@@ -269,45 +251,43 @@ impl Epoch {
                     source: DurationError::Underflow,
                 })
             }
-            Some(years_since_ref) => match years_since_ref.checked_mul(365) {
+            Some(years_since_ref) => match years_since_ref.checked_mul(DAYS_PER_YEAR_NLD as i32) {
```
ChristopherRabotin, line 254: Replace magic number with the constant.

src/epoch/gregorian.rs
```diff
@@ -269,45 +251,43 @@ impl Epoch {
...
             // Add days
-            for year in HIFITIME_REF_YEAR..year {
-                if is_leap_year(year) {
+            for y in HIFITIME_REF_YEAR..year {
+                if is_leap_year(y) {
                     duration_wrt_ref += Unit::Day;
                 }
             }
-            // Remove ref hours from duration to correct for the time scale not starting at midnight
-            // duration_wrt_ref -= Unit::Hour * time_scale.ref_hour() as i64;
```
ChristopherRabotin, line 291: This had been replaced with the time_scale.gregorian_offset()

src/epoch/mod.rs
```diff
@@ -1370,9 +1370,14 @@ fn div_rem_f64(me: f64, rhs: f64) -> (i32, f64) {
 fn div_euclid_f64(lhs: f64, rhs: f64) -> f64 {
     let q = (lhs / rhs).trunc();
     if lhs % rhs < 0.0 {
-        return if rhs > 0.0 { q - 1.0 } else { q + 1.0 };
+        if rhs > 0.0 {
```
ChristopherRabotin, line 1373: Exact same code, just more idiomatic Rust.

src/duration/mod.rs
```diff
@@ -692,7 +690,15 @@ impl fmt::Display for Duration {
             }
 
             let values = [days, hours, minutes, seconds, milli, us, nano];
-            let units = ["days", "h", "min", "s", "ms", "μs", "ns"];
+            let units = [
+                if days > 1 { "days" } else { "day" },
```
ChristopherRabotin, line 694: Super minor but it was driving me crazy that a duration of one day was printed in plural form.

Issue #296, opened 2024-04-24
Optimize the kani formal verification steps


ChristopherRabotin:
As per #289 , it takes over two hours to run the Kani model checker. This task is to optimize this test so we can run it in reasonable time.

ChristopherRabotin:
This ticket will now need to focus on fixing the bugs found in Kani; the CI for Kani now runs once a week.

ChristopherRabotin:
Test `kani_harness_is_gregorian_valid` fails : https://github.com/nyx-space/hifitime/actions/runs/15800867110/job/44538839116

ChristopherRabotin:
Completed in https://github.com/nyx-space/hifitime/pull/481#issue-4344571770 , thanks @feliperodri for all you work !

PR #318, opened 2024-07-22
Kani autogen follow on

ChristopherRabotin:
Linting changes and reducing the kani test surface to get the tests to pass faster.

#296 will deal with fixing these tests.


ChristopherRabotin:
Shucks, Kani still timed out. @cvick32 , off the cuff, would you know which tests I should disable? If not, that's fine, I can look into it later this evening.

Issue #304, opened 2024-05-24
Python submodule to support IDE type hints


ChristopherRabotin:
IDEs don't know what's in the Python module. This can be solved by writing `pyi` files. 

The goal of this task is to essentially repeat what Dora did here: https://github.com/dora-rs/dora/pull/493 .

cardigan1008:
Hi @ChristopherRabotin , I'd like to give it a try. If it's convenient, please assign it to me.

ChristopherRabotin:
With pleasure! I briefly tried the script from the `dora` repo linked above, but it didn't work out of the box, and I don't know why. Let me know if you can figure it out, otherwise, I can probably tackle it in early September according to my current backlog on the three Nyx Space repos.

ChristopherRabotin:
@cardigan1008 , there's a new crate that aims to make this straight forward: https://github.com/PyO3/pyo3/issues/2454#issuecomment-2296314656 .  I have not tried it, but it looks quite promising!

cardigan1008:
> @cardigan1008 , there's a new crate that aims to make this straight forward: [PyO3/pyo3#2454 (comment)](https://github.com/PyO3/pyo3/issues/2454#issuecomment-2296314656) . I have not tried it, but it looks quite promising!

Sounds great! Thanks for reminding. I’ll definitely give it a try soon.

cardigan1008:
Hi @ChristopherRabotin 

Apologize for the delay in updating. I've just completed my project and have had the opportunity to try using pyo3-stub-gen. It turns out that some classes, such as LatestLeapSeconds, LeapSecondsFile, Ut1Provider, PyHifitimeError, PyDurationError, and PyParsingError, work well. However, there are some classes that don’t work as expected, primarily because pyo3-stub-gen fails to parse Duration into its own stub type.

The pyi file generated by pyo3-stub-gen can be found [here](https://gist.github.com/cardigan1008/461767929c0ceb8a87baefb3bba7ba29).

ChristopherRabotin:
Hi there! No problem, this isn't a major rush. Good to know that it works in some cases but not others. Does the stub generator print an error when it fails by any chance? Thanks

cardigan1008:
Yes, when I tried to implement it for Duration. The error goes like this:

```shell
error[E0433]: failed to resolve: could not find `any` in `std`
  --> src/duration/mod.rs:69:32
   |
69 | #[cfg_attr(feature = "python", gen_stub_pyclass)]
[...]

PR #329, opened 2024-09-05
Use pyo3-stub-gen to support IDE type hints

cardigan1008:
For #304 .

This PR introduces the use of `pyo3-stub-gen` to automatically generate `.pyi` files. These stub files provide type hints for Python IDEs.

Currently, the following modules are supported:

- [ ] Epoch
- [x] TimeScale
- [x] TimeSeries
- [ ] Duration
- [x] LatestLeapSeconds
- [x] LeapSecondsFile
- [x] Ut1Provider
- [x] PyHifitimeError
- [x] PyDurationError
- [x] PyParsingError


hifitime.pyi
```diff
@@ -0,0 +1,81 @@
+# This file is automatically generated by pyo3_stub_gen
```
ChristopherRabotin, line 1: It seems that the stub generator could not find any of the functions in any of the classes, so as-is this new `pyi` file won't help IDE users in autocompleting.

cardigan1008: Yes. So maybe we can turn to another way to do this, like dora's?

ChristopherRabotin: I like that dora's approach uses the package introspection, but I could not seem to get it to work correctly when I tried a few weeks (months?) ago. It kept on complaining that one of the return types was not defined although I thought I had. Feel free to give it a try of course!

ChristopherRabotin:
I've implemented Dora's approach in #335 , so I'll close this PR.

Issue #229, opened 2023-05-23
Simplify Epoch initializations from time scales


ChristopherRabotin:
The first PR for version 4 [adds non-exhaustive](https://github.com/nyx-space/hifitime/pull/228) to the `TimeScale` enum. This gives us plenty of flexibility to add additional time scales, like #191. However, we also currently have an `Epoch::from_{time_scale}_[days,seconds,nanoseconds]()` initializer. That's a lot of code, and a good chunk of it is duplication.

With version 4, we have the opportunity to refactor a lot of these initializers.

I recommend that we remove the `Epoch::from_{time_scale}_[days,seconds,nanoseconds]()` initializers and instead allow `std::ops::Mul` between a duration and an time scale for the initialization. The `TimeUnits` trait, included in the `prelude`, allows simple initialization of durations, and so does the `Units` enum.

For example: `Epoch::from_utc_seconds(1234.5)` would become `1234.5.seconds() * TimeScale::UTC`. We could also add a `TimeScaleTrait` to further this initialization as `1234.5.seconds().utc()`. 

The main limitation of this approach is for modified time references like (modified) Julian days: these currently have their own initializers, so we'd need to keep a generic initialization like `Epoch::new(duration: Duration, time_scale: TimeScale, modifier: Option<Modifier>)` where `Modifier` would be Julian, modified Julian, and potentially other ones like the NASA GMAT MJD (which is off by hundreds of years compared to the real MJD).

My slight concern here is that the idea of a "multiplication" between a duration and a time scale does not strictly speaking make any kind of sense. I mostly like the idea of overloading an operator for the initialization instead of requiring the user to use a constructor: this is especially relevant in "script like" usages of the library. Moreover, the multiplication makes it easy to integrate in Python and to keep a similar API: `>>> 1234.5 * Unit.Second * TimeScale.UTC`.

@gwbres , what do you think ?

gwbres:
Hello @ChristopherRabotin,

As far as the code quantity problem goes (related to timescales), the TimeScale trait is the way to go. Constructions like `Epoch::from_seconds(1234.5).utc()` because a Duration would implement the  traits is easy, elegant, and removes 80% of existing methods.

Then, regarding the "Modified" dates, I don't see any problems with keeping existing methods to make things easier.
And if we want to also reduce their code quantity, maybe a dedicated Trait, like `DateModifier`, `Modifier`, _insert any better suited name here_,  might prove more handy too

[...]

PR #332, opened 2024-10-07
MJD/JDE UTC fix + `to_time_scale` now available in Python

### LINKED ISSUE #257: from_gregorian but in GPST
Hello @ChristopherRabotin,

I would be interested in having an `Epoch` construction method similar  to `Epoch::from_gregorian_utc` but in `TimeScale::GPST`.

`from_gregorian_utc` is the best thing to use when all data fields are already known.

Such feature would solve the requirement I face in two of my parsers:

- in RINEX when parsing NAV ephemeris - the SV broadcasts Epoch expressed in their timescale (GPST for example)
- in SP3, all Epochs are expressed in a unique GNSS time scale

### LINKED ISSUE #302: (V4): from_mjd_utc / to_gregorian_utc reciprocal
Hello,

I think I picked up an issue when working with CGGTTS. 
In CGGTTS files, Epochs are expressed in MJD UTC. This is how I parse them:

image

`mjd` being an integer. 
I then use  `to_gregorian_utc`  in the mirror operation: 

image

In my example, I parsed and constructed mjd=59567 h=0 mins=14 seconds=0. 
Then, dumping generates mjd=59567 h=0 mins=14 seconds=37 (leap second issue ?)

### LINKED ISSUE #331: Python lib needs a function to change time scales
This is currently only possible in Rust, but it's quite useful in general, and Python needs to have access to it.

ChristopherRabotin:
Closes #331 
Closes #257 
Closes #302 

This also makes #229 much easier to get started on. My only hesitation with that ticket is that in Python, we use `init_from_unix_seconds` _a lot_ and we're launching in a few weeks so I'm hesitant to change my habits this close to launch.

### DIFF
--- a/src/epoch/initializers.rs
+++ b/src/epoch/initializers.rs
@@ -102,21 +102,18 @@ impl Epoch {
 
     #[must_use]
     pub fn from_mjd_tai(days: f64) -> Self {
+        Self::from_mjd_in_time_scale(days, TimeScale::TAI)
+    }
+
+    pub fn from_mjd_in_time_scale(days: f64, time_scale: TimeScale) -> Self {
         assert!(
             days.is_finite(),
             "Attempted to initialize Epoch with non finite number"
         );
-        Self::from_tai_duration((days - MJD_J1900) * Unit::Day)
-    }
-
-    pub fn from_mjd_in_time_scale(days: f64, time_scale: TimeScale) -> Self {
-        // always refer to TAI/mjd
-        let mut e = Self::from_mjd_tai(days);
-        if time_scale.uses_leap_seconds() {
-            e.duration += e.leap_seconds(true).unwrap_or(0.0) * Unit::Second;
+        Self {
+            duration: (days - MJD_J1900) * Unit::Day,
+            time_scale,
         }
-        e.time_scale = time_scale;
-        e
     }
 
     #[must_use]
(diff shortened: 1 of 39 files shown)


src/epoch/mod.rs
ChristopherRabotin, line 1: Most of the changes here are for the initializers that were moved to `initializers.rs`.

src/epoch/initializers.rs
```diff
@@ -0,0 +1,409 @@
...
+        );
+        Self {
+            duration: (days - MJD_J1900) * Unit::Day,
+            time_scale,
+        }
+    }
+
+    #[must_use]
+    pub fn from_mjd_utc(days: f64) -> Self {
+        Self::from_mjd_in_time_scale(days, TimeScale::UTC)
```
ChristopherRabotin, line 121: This function previously called `from_mjd_tai` and then manually applied the leap seconds. Now it calls the `from_mjd_in_time_scale` function which correctly sets the time scale without changing the duration.

src/epoch/initializers.rs
```diff
@@ -0,0 +1,409 @@
...
+        );
+        Self {
+            duration: (days - MJD_J1900 - MJD_OFFSET) * Unit::Day,
+            time_scale,
+        }
+    }
+
+    #[must_use]
+    pub fn from_jde_utc(days: f64) -> Self {
+        Self::from_jde_in_time_scale(days, TimeScale::UTC)
```
ChristopherRabotin, line 158: The `from_jde_utc` function has the same bug as in #302 , so I've done the same fix.

tests/epoch.rs
```diff
@@ -2069,11 +2069,17 @@ fn regression_test_gh_282() {
 fn regression_test_gh_288() {
     use core::str::FromStr;
     let epoch = Epoch::from_str("2021-03-06 11:14:40.9960 GPST").unwrap();
+    let epoch_from_gpst =
```
ChristopherRabotin, line 2072: Additional test for #257 .

tests/epoch.rs
```diff
@@ -2137,3 +2143,32 @@ fn regression_test_gh_317() {
         nanos
     );
 }
+
+#[test]
+fn regression_test_gh_302() {
```
ChristopherRabotin, line 2148: New regression test for #302 .

tests/python/test_epoch.py
```diff
@@ -25,6 +25,9 @@ def test_strtime():
         print(f"caught {e}")
     else:
         raise AssertionError("failed to catch parsing error")
+    
+    epoch_tdb = epoch.to_time_scale(TimeScale.TDB)
```
ChristopherRabotin, line 29: Add a Python test for #331.

Issue #337, opened 2024-10-18
Conversions to/from `std::time::Duration` are lossy and easy to misuse


max-heller:
The implementations of `From<std::time::Duration> for Duration` and `From<Duration> for std::time::Duration` are lossy because
- They go from ints to floats to ints instead of from int to int (both `Duration`s are based on an integer number of nanoseconds)
  - This *is not* documented
- They saturate at `Duration::ZERO` and `Duration::MAX`
  - This *is* documented

https://github.com/nyx-space/hifitime/blob/cdcbc9f34b5a243eafcd7bbdebc10523b743fb07/src/duration/std.rs#L17-L45

The docs for `From` [suggest that implementations should be lossless](https://doc.rust-lang.org/std/convert/trait.From.html#when-to-implement-from) and I've run into difficult-to-track down bugs because of the lossiness and ease of calling `From`/`Into`. I see a couple of options that would make more sense:
1. Keep `From` implementations with saturating behavior, but remove the lossiness caused by the float conversion. This is still confusing, but much less so than running into float precision losses.
2. Replace the `From` implementations with `TryFrom` implementations to account for possible overflow, and also use integer conversions instead of float conversions

ChristopherRabotin:
Hi Max,

You raise a good point: a number of the initializers in hifitime are marked as error-less because the representation is manually minimized or maxed out.

I've just released version 4.0.0. Ideas for version 5 are already being worked on. Switching from the error-free `From` trait to the `TryFrom` trait should be one of the improvements for version 5.

In the meantime, version 4.1 could implement the first solution you propose by getting rid of the integer conversions.

ChristopherRabotin:
Fixed in #342 by @Wollaston .

PR #342, opened 2024-10-19
Implement lossless From methods for Duration conversions

Wollaston:
This works towards closing #337 by making the implementations of `From<std::time::Duration> for Duration` and `From<Duration> for std::time::Duration` lossless. Thank you!

### DIFF
--- a/src/duration/std.rs
+++ b/src/duration/std.rs
@@ -12,25 +12,29 @@
 
 extern crate core;
 
-use super::{Duration, Unit};
+use super::Duration;
 
 impl From<Duration> for std::time::Duration {
     /// Converts a duration into an std::time::Duration
     ///
     /// # Limitations
     /// 1. If the duration is negative, this will return a std::time::Duration::ZERO.
-    /// 2. If the duration larger than the MAX duration, this will return std::time::Duration::MAX
+    /// 2. If the duration is larger than the MAX duration, this will return std::time::Duration::MAX
     fn from(hf_duration: Duration) -> Self {
-        const NANOS_PER_SEC: u128 = std::time::Duration::from_secs(1).as_nanos();
-        let nanos = hf_duration.total_nanoseconds();
-        let unsigned_nanos = u128::try_from(nanos).unwrap_or(0);
+        use crate::NANOSECONDS_PER_SECOND;
+        if hf_duration.signum() == -1 {
+            std::time::Duration::ZERO
+        } else {
+            let nanos = hf_duration.total_nanoseconds();
+            let unsigned_nanos = u128::try_from(nanos).unwrap_or(0);
 
-        let secs: u64 = (unsigned_nanos / NANOS_PER_SEC)
-            .try_into()
-            .unwrap_or(u64::MAX);
-        let subsec_nanos = (unsigned_nanos % NANOS_PER_SEC) as u32;
+            let secs: u64 = (unsigned_nanos / NANOSECONDS_PER_SECOND as u128)
+                .try_into()
+                .unwrap_or(u64::MAX);
+            let subsec_nanos = (unsigned_nanos % NANOSECONDS_PER_SECOND as u128) as u32;
 
-        std::time::Duration::new(secs, subsec_nanos)
+            std::time::Duration::new(secs, subsec_nanos)
+        }
     }
 }
 
@@ -39,7 +43,7 @@ impl From<std::time::Duration> for Duration {
     ///
     /// # Limitations
     /// 1. If the duration is negative, this will return a std::time::Duration::ZERO.
-    /// 2. If the duration larger than the MAX duration, this will return std::time::Duration::MAX
+    /// 2. If the duration is larger than the MAX duration, this will return std::time::Duration::MAX
     fn from(std_duration: std::time::Duration) -> Self {
         Duration::from_total_nanoseconds(std_duration.as_nanos().try_into().unwrap_or(i128::MAX))
     }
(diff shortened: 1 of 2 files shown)


src/duration/std.rs
```diff
@@ -21,15 +21,16 @@ impl From<Duration> for std::time::Duration {
...
-        let (sign, days, hours, minutes, seconds, milli, us, nano) = hf_duration.decompose();
-        if sign < 0 {
-            std::time::Duration::ZERO
-        } else {
-            // Build the seconds separately from the nanos.
-            let above_ns_f64: f64 =
-                Duration::compose(sign, days, hours, minutes, seconds, milli, us, 0).to_seconds();
-            std::time::Duration::new(above_ns_f64 as u64, nano as u32)
-        }
+        const NANOS_PER_SEC: u128 = std::time::Duration::from_secs(1).as_nanos();
```
ChristopherRabotin, line 24: I think a simpler way to implement this functionality is by using the [`to_unit`](https://docs.rs/hifitime/latest/hifitime/struct.Duration.html#method.to_unit) function on the `hf_duration`.

I haven't tested the following, so consider this a draft.

```rust
let seconds = hf_duration.to_seconds().floor() as u64; // Somehow check that this won't fail
let subsec_nanos = (hf_duration - Unit::Second * seconds).to_unit(Unit::Nanoseconds).floor() as u32; // Also check that this won't fail, probably using try_from as you initially recommended

std::time::Duration(seconds, subsec_nanos)
```

ChristopherRabotin:
Thanks for your PR @Wollaston ! I've proposed a simpler way to have a lossless conversion to std::time::Duration: it's lossless because the operations on `Duration` are lossless, including the unit conversions. The first `floor` subtracts anything above one second from the duration, thereby ensuring that the `to_unit(Unit::Nanosecond)` will only be seconds.

Is there a way we could test this?

src/duration/std.rs
ChristopherRabotin: There's a better way to handle the `subsec_nanos` than what I wrote above. After the subtraction, we know that the duration has zero centuries. So, we can safely ignore that part and operate directly on nanoseconds.

```rust
use crate::NANOSECONDS_PER_SECOND; // https://docs.rs/hifitime/latest/hifitime/constant.NANOSECONDS_PER_SECOND.html
let subsec_nanos: u64 = (hf_duration - Unit::Second * seconds).nanoseconds / NANOSECONDS_PER_SECOND;
// Finally convert this to u32 safely.
```

Wollaston:
Thanks for the review @ChristopherRabotin! In the original issue #337, part of the concern was the conversion from ints to floats to ints, which is lossy with the float conversion. In your draft, `hf_duration.to_seconds()` does do a float conversion, so my inclination is to stay with the original PR that avoids floats altogether. 

As to the updated `subsec_nanos` calculation, I was not able to get that to compile due to the following error: "cannot multiply `timeunits::Unit` by `u64` the trait `std::ops::Mul<u64>` is not implemented for `timeunits::Unit`". Because of that, perhaps it would be simpler overall to keep with the original calculation as it avoids any compile issues, or needing to implement behavior elsewhere.

I agree it makes sense to use the `crate::NANOSECONDS_PER_SECOND` instead of the one calculated locally here.

As to tests, I can draft some unit tests in the module (it only has these two implementations) and/or add some doc tests to some updated documentation for these implementations. I am not sure if there is a pattern you prefer - just let me know on that. They could test that the updated implementations are in fact lossless as a start. 

I can update the PR accordingly once those questions are settled, and then squash the commits. Many thanks!

ChristopherRabotin:
You're right. Let's go ahead with your original implementation using the crate's NANOSECONDS_PER_SECOND constant. As for negative durations, where we'd want to return `std::time::Duration::ZERO`, there's now a `const fn` [`signum`](https://docs.rs/hifitime/latest/hifitime/struct.Duration.html#method.signum) function which I recommend using here.

Wollaston:
Sounds good! Just to clarify, we still want to return `std::time::Duration::ZERO` in the case of negative durations, correct? I might be overlooking how to use `const fn signum` for these implementations. The use of `u128::try_from(nanos).unwrap_or(0)` seems to catch the cases of negative durations without needing to use `const fn signum`, so maybe I am missing the best way to use that. 

And for the tests, do you have a preference in how the tests are added, or just that they are? Thanks!

ChristopherRabotin:
Yes, I think it makes sense to return `Duration::ZERO` when the hifitime duration is negative. The advantage of using the const fn is that, I believe, it's generally faster to evaluate than the [`unwrap_or`](https://doc.rust-lang.org/std/result/enum.Result.html#method.unwrap_or), even though functionally it should be the same I agree.

By the way, no need to squash the commits, I tend to just merge all of the commits.

ChristopherRabotin:
Sorry, this PR fell off my radar. I'll try to release version `4.0.1` this weekend for a minor bug fix in the Python type hints. I was wondering whether I could also merge your work and release it then.

Thanks

Wollaston:
Good morning Chris - I also got pulled away for a few things, but I am planning on writing tests today. That should be the last step for this PR, barring any feedback. The current branch includes the `signum` function as a first check/short-circuit, before proceeding with the rest of the logic. It's linked in this thread. If you had something else in mind, just let me know.

As to tests, I am adding some big number tests to show this update is lossless. I saw that you already test negative duration and small intervals. So I will add tests with large numbers and saturation. Thanks again.

ChristopherRabotin:
Thanks Matt, this looks great! I'll go ahead and merge.

Issue #359, opened 2025-04-02
Replace reqwest dependency with a lighter alternative: ureq


FusRoman:
Hi,

First of all, thanks for the great work on hifitime — I'm using it for an academic project and it's been really helpful.

While optimizing the dependencies of my project, I noticed that enabling the ut1 feature significantly increases the number of dependencies due to the use of the reqwest crate. Since hifitime only uses the blocking API of reqwest, I’d like to suggest switching to a lighter alternative like ureq, which provides similar functionality in a blocking context but with a much smaller dependency footprint.

This change would help reduce build times and make the dependency tree cleaner when using the ut1 feature.

Let me know what you think — I’d be happy to help with a PR if you're open to the idea.

ChristopherRabotin:
Done in https://github.com/nyx-space/hifitime/pull/367.

PR #360, opened 2025-04-03
359 replace reqwest with ureq

FusRoman:
Hi,

This PR is related to issue #359 and replaces the reqwest dependency with ureq, a lighter HTTP client crate.

The overall code change is minimal, as ureq and reqwest share a very similar API. Although I didn’t perform any detailed timing benchmarks, the execution should now be slightly faster since the tokio runtime is no longer being initialized.

Additionally, I fixed the UT1 test related to the JPL request. After switching to ureq, the test started returning a 403 Forbidden error for the JPL file version 221222_190002. It turned out that the file had been moved into a 2022/ subdirectory. I’m not sure why this didn’t cause an issue previously with reqwest. In any case, I updated the test URL to include the correct subdirectory path (2022/) in order to resolve the problem. The test now performs the correct conversion, consistent with the corresponding Astropy-based comment.

Let me know if anything needs adjusting!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,17 +35,19 @@ lexical-core = { version = "1.0.1", default-features = false, features = [
     "parse-integers",
     "parse-floats",
 ] }
-reqwest = { version = "0.12", features = ["blocking", "json"], optional = true }
 tabled = { version = "0.17.0", optional = true }
 openssl = { version = "0.10", features = ["vendored"], optional = true }
 web-time = { version = "1.0.0", optional = true }
 snafu = { version = "0.8.2", default-features = false }
+ureq = { version = "3.0.10", default-features = false, optional = true, features = [
+    "rustls",
+] }
 
 [features]
 default = ["std"]
 std = ["serde", "serde_derive", "web-time", "snafu/std", "snafu/backtrace"]
 python = ["std", "pyo3", "ut1"]
-ut1 = ["std", "reqwest", "tabled", "openssl"]
+ut1 = ["std", "ureq", "tabled", "openssl"]
 
 [dev-dependencies]
 serde_json = "1.0.91"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -81,6 +81,9 @@ web-sys = { version = "0.3", features = [
     'PerformanceTiming',
 ] }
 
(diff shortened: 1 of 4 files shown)


src/epoch/ut1.rs
```diff
@@ -88,18 +89,19 @@ impl Ut1Provider {
...
-            }
-            Err(e) => Err(HifitimeError::Parse {
-                source: ParsingError::DownloadError {
-                    code: e.status().unwrap_or(StatusCode::SEE_OTHER),
-                },
+        let url = format!("https://eop2-external.jpl.nasa.gov/eop2/{}", version);
+
+        // Download the file
+        match get(url).call() {
+            Ok(resp) => Self::from_eop_data(resp.into_body().read_to_string().expect(
```
ChristopherRabotin, line 96: Although the previous code used an `unwrap`, and `expect` is definitely better than `unwrap`, I think it's best to return a Parse UnknownFormat if the `read_to_string` fails.

ChristopherRabotin (approved):
Thanks for the PR! I'm looking into the action failing, but I wonder linux aarch64 should just be dropped.

Edit: In ANISE, aarch64 is a supported platform. I'm not sure why the `ring` crate is causing problems.

Cargo.toml
ChristopherRabotin, line 1: Can you try to add 

```toml
[target.aarch64-unknown-linux-gnu.env]
CFLAGS="-D__ARM_ARCH=8"
```

So we can check if that fixes the aarch64 build.

src/epoch/ut1.rs
FusRoman: Definitely, I did that on my last commit.

Cargo.toml
FusRoman: added in the last commit

ChristopherRabotin: Thanks for the info. Could you edit line 36 of `.github/workflows/tests.yml` to be `- { version: "1.81", name: MSRV }`: this will bump up the MSRV to 1.81, which was released in September 2024. I think it's reasonable to move up the MSRV.

(I can't seem to push to your fork, which is why I'm asking for you the changes.)

Thanks

ChristopherRabotin:
Hey, I've cloned your branch and updated the PR as #367 .

Issue #291, opened 2024-04-03
Lunar coordinated time 


ChristopherRabotin:
By the end of 2026, NASA should have a formal definition of LCT. Hifitime should provide some example reference to support that.

> OSTP chief Arati Prabhakar's memo said that for a person on the moon, an Earth-based clock would appear to lose on average 58.7 microseconds per Earth-day and come with other periodic variations that would further drift moon time from Earth time. 

https://www.reuters.com/science/white-house-directs-nasa-create-time-standard-moon-2024-04-02/

mr5x5x5:
I had fun thinking about this all day today. Well, relatively all-day.
Also [A Relativistic Framework to Establish Coordinate Time on the Moon and Beyond
by Neil Ashby, Bijunath Patla](https://arxiv.org/abs/2402.11150)

ChristopherRabotin:
Nice ! And I see you've implemented it in Python [here](https://github.com/mr5x5x5/MoonOclock). @philiplinden also worked with time dilation [here](https://github.com/philiplinden/spacetime).

If LCT is to work similarly to the TDB or ET time systems, then it'll need a reference epoch from which the drift happens continuously. If that reference is set to 01 Jan 2000 at noon, then we'd have accumulated 520 ms 62 μs difference with Earth time so far in these 8859 days.

Calculations for reference:
```
>>> from anise.time import *
>>> lct_delta = Unit.Microsecond*58.7
>>> num_days = (Epoch.system_now().timedelta(Epoch("2000-01-01 12:00:00 TDB"))).to_unit(Unit.Day)
>>> lct_delta * num_days
520 ms 33 μs 92 ns @ 0x253b19525c0
>>> num_days
8859.166822171665
```

jmfriedt:
If it can be of interest, I have just uploaded Frederic Meynadier (BIPM) and Pascale Defraigne (Obs. Brussels) presentation at European Frequency and Time Forum (EFTF) this June 2024 on the topic at http://jmfriedt.free.fr/EFTF2024_moon.mp4 
The video is of poor quality and the slides hardly readable but I can ask for a copy if interested.

ChristopherRabotin:
Superb, thank you! I would be very interested in seeing the slides too if you can ask for a copy. Thanks.

philiplinden:
[...]

PR #361, opened 2025-04-13
TimeOffset definition for precise TimeScale conversions

gwbres:
Timescale monitoring, comparison to others, maintenance and precise navigation in all timescales require precise conversion between two of them. Currently we have `Epoch::to_time_scale` which is not a precise conversion because it has no means to take into account the state of the reference timescale at the epoch of conversion. 

A precise conversion is a short term interpolation by using polynomials terms that are published regularly (for example by Clock RINEX files for timescale mainteners, or NAV RINEX files for GNSS constellations), and expressed in the current week of the reference timescale. That's the main reason why I already introduced `Epoch::to_time_of_week` and `Epoch::from_time_of_week`, which offer means to describe the Reference epoch of this operation or Keplerian navigation. For precise work, the interpolation interval should remain the shortest possible. it is up to the user to keep its TimeOffset structure up to date. I arbitrarly set an upper limit to a week long duration: the TimeOffset will not allow precise conversion between two timescales, if the conversion parameters are more than a week old.

When we say `|GPST-UTC|` it means this timeoffset allows converting GPST to UTC precisely, and UTC to GPST precisely by reversing (`-`) the operation. The TimeOffset object is smart enough to support the converion both ways: the user can arrive with either GPST or UTC in that particular example. 

I think it would be great to see my `TimeOffset` structure integreted into Hifitime, this way, Hifitime would not only define all Timescales we can work with, but allow precise conversions between them and provide all the basic tools for people involved in Timescale maintenance & monitoring.

### DIFF
--- a/src/timeoffset.rs
+++ b/src/timeoffset.rs
@@ -94,6 +94,15 @@ impl TimeOffset {
             .unwrap()
     }
 
+    /// Define a new |[TimeScale::BDT] - [TimeScale::GPST]| [TimeOffset] from time of week components
+    pub fn from_bdt_gpst_time_of_week(
+        t_ref_nanos: (u32, u64),
+        polynomials: (f64, f64, f64),
+    ) -> Self {
+        Self::from_reference_time_of_week(t_ref_nanos, TimeScale::BDT, TimeScale::GPST, polynomials)
+            .unwrap()
+    }
+
     /// Define a new |[TimeScale::BDT] - [TimeScale::UTC]| [TimeOffset] from time of week components
     pub fn from_bdt_utc_time_of_week(
         t_ref_nanos: (u32, u64),
--- a/src/timeoffset.rs
+++ b/src/timeoffset.rs
@@ -229,9 +229,9 @@ impl TimeOffset {
 
         // support back & forth conversion
         if t.time_scale == self.rhs {
-            Ok(-dt_s)
+            Ok(-dt_s * 1.0E9)
         } else {
-            Ok(dt_s)
+            Ok(dt_s * 1.0E9)
         }
     }
 
@@ -265,7 +265,7 @@ mod test {
     #[test]
     fn test_1ns_time_offset() {
         // Tests the TimeOffset API with values slightly above hifitime precision.
-        let polynomials = (1E-9, 1E-9, 0.0);
+        let polynomials = (1E-9, 0.0, 0.0);
 
         let known_timescales = [
             TimeScale::UTC,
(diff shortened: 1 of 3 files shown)


ChristopherRabotin:
I like the idea of precise time-variable time scales. That would also enable both definitions of #291. Let me review the code, and in the meantime, could you bump up the minor version of hifitime and the minimum supported Rust version since that seems to be required now sadly.

src/timeoffset.rs
```diff
@@ -0,0 +1,362 @@
...
+    IdenticalTimescales,
+    /// Epoch is not expressed in one of the two supported [TimeScale]s!
+    NotSupportedTimescale,
+    /// Epoch should be expressed in the left-hand side [TimeScale].
+    InvalidTimescale,
+    /// [TimeOffset] is outdated and should not apply (difference is too large): should be updated weekly at least!
+    OutdatedTimeOffset,
+}
+
+/// [TimeOffset] used in [TimeShift]ing operations
```
ChristopherRabotin, line 16: I think this documentation should provide an example of when this is useful, like in your PR description with the UTC-GPST difference.

The documentation references the `TimeShift` struct, but that isn't in the PR.

src/timeoffset.rs
```diff
@@ -0,0 +1,362 @@
...
+#[derive(Copy, Clone, PartialEq)]
+pub struct TimeOffset {
+    /// Left hand side (compared) [TimeScale]
+    lhs: TimeScale,
```
ChristopherRabotin, line 20: Maybe rename `lhs` and `rhs` to `compared_lhs` and `reference_rhs`?

src/timeoffset.rs
```diff
@@ -0,0 +1,362 @@
...
+    t_ref_nanos: (u32, u64),
+    /// Polynomials (s, s.s⁻¹, s.s⁻²)
+    polynomials: (f64, f64, f64),
+}
+
+impl TimeOffset {
+    /// Define a new [TimeOffset] from this reference [Epoch] expressed
+    /// in left-hand side [TimeScale] to rhs [TimeScale]
+    pub fn from_reference_epoch(
+        t_ref: Epoch,
```
ChristopherRabotin, line 33:
```suggestion
        ref_epoch: Epoch,
```
That's the nomenclature used in a few parts of the code.

src/timeoffset.rs
```diff
@@ -0,0 +1,362 @@
...
+        rhs: TimeScale,
+        polynomials: (f64, f64, f64),
```
ChristopherRabotin, line 35: I think `polynomials` should be its own structure, and be `Duration`s instead of an `f64` of seconds.

```rust
pub struct Polynomial {
    constant: Duration,
    rate: Duration,
    accel: Duration
}
```

This would allow the definition of complicated offsets, and should actually allow the redefinition of the currently supported time scales.

src/timeoffset.rs
```diff
@@ -0,0 +1,362 @@
...
+
+/// [TimeOffset] used in [TimeShift]ing operations
+#[derive(Copy, Clone, PartialEq)]
+pub struct TimeOffset {
+    /// Left hand side (compared) [TimeScale]
+    lhs: TimeScale,
+    /// Right hand side (reference) [TimeScale]
+    rhs: TimeScale,
+    /// Weekly reference time (counter, nanoseconds)
+    t_ref_nanos: (u32, u64),
```
ChristopherRabotin, line 24: This seems a bit too specific to GPST, no? Weekly reference and nanoseconds. Can we make this more generic somehow?

src/timeoffset.rs
```diff
@@ -0,0 +1,362 @@
...
+        Ok(Self {
+            lhs,
+            rhs,
+            t_ref_nanos,
+            polynomials,
+        })
+    }
+
+    /// Define a new |[TimeScale::GPST] - [TimeScale::UTC]| [TimeOffset] from time of week components
+    pub fn from_gpst_utc_time_of_week(
```
ChristopherRabotin, line 71: In the original API, we had a lot of initializers for all of the different time scales: let's avoid this mistake again, it was difficult to remove once added. Let's expect the users to use the generic initializer and correctly build their own epoch before calling it.

gwbres:
> and in the meantime

I'm also in the process of adding more tests. In particular, because this operation works both ways, so mirroring should always work and needs to be tested. To make things clearer, a let's take the GPST/UTC TimeOffset notated `|GPST(=lhs)-UTC=(rhs)|`,  allows precise GPST->UTC conversion but also UTC->GPST conversion by using the backwards equation.

> could you bump up the minor version of hifitime and the minimum supported Rust version since that seems to be required now sadly

not sure what you mean by that ? my branches are always aligned to your latest master

src/timeoffset.rs
```diff
@@ -0,0 +1,362 @@
...
+    /// NB: this should be expressed in the left-hand side [TimeScale] and we have no means
+    /// to verify that.
+    pub fn update_mut(&mut self, t_ref_nanos: (u32, u64), polynomials: (f64, f64, f64)) {
+        self.t_ref_nanos = t_ref_nanos;
+        self.polynomials = polynomials;
+    }
+
+    /// Define a new [TimeOffset] with new reference [TimeScale], while preserving other components.
+    /// NB: this should be expressed in the left-hand side [TimeScale] and we do not verify it!
+    pub fn with_reference_timescale(&self, ts: TimeScale) -> Self {
```
ChristopherRabotin, line 167:
```suggestion
    pub fn with_reference_timescale(mut self, ts: TimeScale) -> Self {
        self.rhs = ts;
        self
    }
```

I learned only recently that for `Copy` structs, you can just move the structure with the function call, and avoid calling `clone`.

src/timeoffset.rs
```diff
@@ -0,0 +1,362 @@
...
+        let mut s = self.clone();
+        s.rhs = ts;
+        s
+    }
+
+    /// Define a new [TimeOffset] with new left-hand side [TimeScale], while preserving other components.
+    /// This needs to be coupled to either [Self::with_reference_time_of_week_nanos] or
+    /// [Self::with_reference_epoch] to remain correct.
+    pub fn with_lhs_timescale(&self, ts: TimeScale) -> Self {
```
ChristopherRabotin, line 176: Idem on all `(&self ... ) -> Self` functions.

src/timeoffset.rs
```diff
@@ -0,0 +1,362 @@
...
+        s.polynomials = polynomials;
+        s
+    }
+
+    /// Returns the total number of nanoseconds to apply to convert this [Epoch]
+    /// into either of [Self::supported_timescales].
+    /// NB:
+    /// - `t` must be expressed in either of [Self::supported_timescales].
+    /// - `t` should fall within the reference week, otherwise this will give invalid results.
+    pub fn time_correction_nanos(&self, t: Epoch) -> Result<f64, TimeOffsetError> {
```
ChristopherRabotin, line 213: If `polynomials` use `Duration`, then most of this math becomes much simpler since `Duration` already handles the math.

src/timeoffset.rs
gwbres: About the TimeShift stuff, sorry I did not intend to join it to this PR. It's part of my own stuff. I think it's up to the user to do whatever they have, while Hifitime remains the building blocs

src/timeoffset.rs
gwbres: Not it is totally generic, what makes you say that ?

It's just the calculations (the interpolation) uses time of week and week referencing, not timescale::t_0 referencing. 
And we have means to express weeks and reference to a certain week using your library

src/timeoffset.rs
gwbres: So Duration would only apply to the first term. And IMO it is important to not use Durations here, because we allow the user to not be limited to 1ns precision.

src/timeoffset.rs
gwbres: I tried to apply this everywhere. Does that buy you a memcopy ?

gwbres:
I am also up to 

- rename `TimeOffset` to whatever name that makes more sense to you (if need be).
- move tests to whatever you prefer
- discuss the newly introduced Error. For me it's important to either provide an error related to these operations, or simply panic. Because it is kind of easy to obtain invalid results if the user starts tweaking things (like redefinition of either timescale). I'm up to rename this error or put it in your dedicated folder

gwbres:
Maybe TimeOffset is badly named. The way I see this now, is that `Epoch.to_time_scale` is a coarse correction, while this would provide the fine correction you add on top of that, to obtain the true state of the timescale (at a specific point in time)

src/timeoffset.rs
ChristopherRabotin: Makes sense

src/timeoffset.rs
ChristopherRabotin: I thought of GPST because I've never seen the start of week as something important in epochs outsides of GPST. Some astrodynamics data references day of year, but that's also pretty specific.

src/timeoffset.rs
ChristopherRabotin: In Nyx, I define a [`CommonPolynomial`](https://docs.rs/nyx-space/latest/nyx_space/polyfit/enum.CommonPolynomial.html) for maneuver definitions. I see a similarity here, which is why I proposed a polynomial structure.

gwbres:
I'm proposing another approach in the other PR. Feel free to close either (or both :D) when we converge to a conclusion.

gwbres:
Well I guess this should be closed then. I just hope the 1ns accuracy limitation does not prove limitating too quickly

PR #365, opened 2025-04-21
Polynomial: python + formatter

gwbres:
Provides `Display` for the new Structure.
Provides python basics for new structure.


gwbres:
Since this is very small, I did not bother split it into a folder, just like `TimeSeries`

ChristopherRabotin (approved):
Looks good!

gwbres:
> Looks good! Go ahead and merge once the tests pass.

I'll be back in a few hours

ChristopherRabotin:
Hi Guillaume,

How have the tests been going? I've now exported the Polynomial class to the Python module in #367 and merged other dependencies. If your tests have been successful, can we release version 4.1.0 soon? If not, that's OK, I can release version 0.5.4 of ANISE without hifitime 4.1.0 as well.

Thanks

gwbres:
Hello, if you release a new ANISE then it's better to have it updated as well.
On my side, I demonstrated that the operation seems to do its job, I can have measurements expressed in other timescales and not still obtain the same geometry. For the absolute time being solved, I don't have a great use case because the timescales are very close to one another and the 1ns accuracy limits us.
I also have not updated RINEX to this day, for one reason, is the idea to encode the polynomial as Duration limits the accuracy, so anything that comes out of RINEX parsing, is simply 0. I'm considering making the upgrade anyway, and it will naturally update once we improve that. Bottom line, the 1ns limitation for this topic is very problematic

ChristopherRabotin:
Okay. If the 1 ns accuracy is a limitation, then I prefer that you make the necessary changes in the Polynomial struct to fix this limitation, and then that we release version 4.1.0. I think I better understand the issue now with the Duration for the inputs to Polynomial, so I am happy with you changing this back to a polynomial of nanoseconds.

gwbres:
I need more time to conclude my current devs, once completed I will run a few more tests, mostly re-run my general test that mixes all constellation/timescale cases.

I'm not sure that breaking the polynomial struct is right thing to do nor even needed. The application logic is fine in any case

ChristopherRabotin:
Okay, I'm not in a hurry, so take your time to make sure the design fixes your use case.

gwbres:
Currently the "only problem" I face with the current master, is that the precision is often too low to describe typical timescale specifications, so the RINEX Parser winds up with (0, 0, 0) = null correction quite often.
When it is not null, the correction is correct and it seems to work just fine ! 
If you agree, I'll push back to (f64, f64, f64) internal triplet, so I can confirm the update of all my stuff to this new object

gwbres:
I don't know, in the end that's an Epoch we want to translate and stirr so the 1ns limitation is always there.
And until that is changed, I don't think that changing your library to accomodate mines is a good idea.
I'll just have my (f64, f64, f64) wrapper in RINEX, and content of the 1ns limitation in navigation applications until further notice

PR #369, opened 2025-04-26
Polynomial: provide convenient (f64, f64, f64) casts

gwbres:
- Provide casts that might prove convenient to the end users

### DIFF
--- a/src/polynomial.rs
+++ b/src/polynomial.rs
@@ -40,6 +40,31 @@ pub struct Polynomial {
     pub accel: Duration,
 }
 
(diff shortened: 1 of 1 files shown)


ChristopherRabotin:
Will the change to nanosecondsbbe in another PR?

ChristopherRabotin (approved):

gwbres:
> Will the change to nanosecondsbbe in another PR?

I don't think we need this ! I already have means to parse the RINEX triplets, and its OK for the Epoch stirring method to be limited - as long as we're aware of it and we announce it to the end users

PR #372, opened 2025-04-27
Polynomial: add two other tests

gwbres:
* static offset (a0, 0, 0) model
   * static offset + linear (a0, a1, 0) model

### DIFF
--- a/tests/polynomial.rs
+++ b/tests/polynomial.rs
@@ -1,0 +1,238 @@
+use hifitime::{
+    prelude::{Duration, Epoch, TimeScale, Unit},
+    Polynomial,
+};
+
+use std::str::FromStr;
+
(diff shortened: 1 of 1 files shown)


gwbres:
I managed to write two other thorough tests, that have the benefit to be easy to be fuzzed around.

One test for the (a0, 0, 0) easiest model.
One test for the (a0, a1, 0) more complex model.

Both tests verify positive and negative offsets (the two possible use cases),  
forward and backwards stirring and reciprocity of the operation.

The principal is "simple", it you take a starting point datetime, and apply a duration offset: that will be your interpolation instant.

ChristopherRabotin:
Looks good, approved. Feel free to merge this one and the other PR whenever.

PR #374, opened 2025-04-29
Correct UTC import in Python unit tests

emmanuel-ferdman:
# PR Summary
This PR fixes a bug where `datetime.UTC` caused an `AttributeError` in Python versions where `UTC` is not a class attribute. The error occurred because `UTC` is accessed via the `datetime` module (`datetime.UTC`), not the `datetime.datetime` class. Replaced `datetime.UTC` with the widely compatible `timezone.utc` to ensure cross-version compatibility (we could import `UTC` from `datetime` instead). Also fixed a small typo along the way.

### DIFF
--- a/tests/python/test_epoch.py
+++ b/tests/python/test_epoch.py
@@ -36,7 +36,7 @@ def test_strtime():
 def test_utcnow():
     epoch = Epoch.system_now()
     try:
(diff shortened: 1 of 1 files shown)


ChristopherRabotin:
Sorry, I totally forgot about this PR! Looks good, I'll merge it right now.

PR #379, opened 2025-05-20
Add Epoch::from_ptp_*

### LINKED ISSUE #378: Support PTP TAI epoch
PTP [uses](https://en.m.wikipedia.org/wiki/Precision_Time_Protocol#Architecture) the unix epoch but the TAI timescale, and is widely used in geospatial survey equipment like GNSS receivers, LiDAR scanners, cameras etc.

Currently, `Epoch::from_tai_seconds` (and similar) assume the 1900 epoch. It would be great if either the epoch could be specified (best option) or if there would be a `Epoch::from_ptp_seconds()` method (or similar).

I’m willing to submit a PR, but since this makes (reference) epochs configurable, I figured I’ll open an issue first.

georgeboot:
Fixes #378

### DIFF
--- a/src/epoch/initializers.rs
+++ b/src/epoch/initializers.rs
@@ -342,6 +342,20 @@ impl Epoch {
     }
 
     #[must_use]
+    /// Initialize an Epoch from the provided IEEE 1588-2008 (PTPv2) second timestamp since TAI midnight 1970 January 01.
+    /// PTP uses the TAI timescale but with the Unix Epoch for compatibility with unix systems.
+    pub fn from_ptp_seconds(seconds: f64) -> Self {
+        Self::from_tai_duration(UNIX_REF_EPOCH.to_tai_duration() + seconds * Unit::Second)
+    }
+
+    #[must_use]
+    /// Initialize an Epoch from the provided IEEE 1588-2008 (PTPv2) millisecond timestamp since TAI midnight 1970 January 01.
+    /// PTP uses the TAI timescale but with the Unix Epoch for compatibility with unix systems.
+    pub fn from_ptp_milliseconds(millisecond: f64) -> Self {
+        Self::from_tai_duration(UNIX_REF_EPOCH.to_tai_duration() + millisecond * Unit::Millisecond)
+    }
+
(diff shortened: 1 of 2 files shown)


ChristopherRabotin:
Thanks for the PR! Could you also add a from_ptp_duration constructor, similar to https://docs.rs/hifitime/latest/hifitime/struct.Epoch.html#method.from_tai_duration.

Thanks

ChristopherRabotin:
Hmm, I can tackle the dependency error. It seems like the MSRV may need updating again.

src/epoch/initializers.rs
```diff
@@ -341,6 +341,20 @@ impl Epoch {
...
+    /// Initialize an Epoch from the provided IEEE 1588-2008 (PTPv2) second timestamp since TAI midnight 1970 January 01.
+    /// PTP uses the TAI timescale but with the Unix Epoch for compatibility with unix systems.
+    pub fn from_ptp_seconds(seconds: f64) -> Self {
+        Self::from_tai_duration(UNIX_REF_EPOCH.to_tai_duration() + seconds * Unit::Second)
+    }
+
+    #[must_use]
+    /// Initialize an Epoch from the provided IEEE 1588-2008 (PTPv2) millisecond timestamp since TAI midnight 1970 January 01.
+    /// PTP uses the TAI timescale but with the Unix Epoch for compatibility with unix systems.
+    pub fn from_ptp_milliseconds(millisecond: f64) -> Self {
```
gwbres, line 354: Minor thought,

(I also like having construction methods like these a lot, they make life much easier for the end user)
but since PTP applications are usually located around tens of nanoseconds (usually) I would rather have `from_ptp_nanoseconds` instead of milliseconds ?

georgeboot:
> Hmm, I can tackle the dependency error. It seems like the MSRV may need updating again.

Yeah not too sure what to do about that.

ChristopherRabotin:
Thanks @georgeboot , I'll merge this now. I'm working on fixing the lints on another branch. It'll be released with version 4.1.0 in the coming days.

PR #410, opened 2025-08-01
Check max days in feb month during a leap year

ephraim71:
This will now give out a invalid georgian date : 
```
    let e = Epoch::from_gregorian_tai(2024, 2, 30, 00, 00, 0,0);
    println!("{e}");
```

Previously, Feb 30 and Feb 31 in an leap year were considered valid.

### DIFF
--- a/src/epoch/gregorian.rs
+++ b/src/epoch/gregorian.rs
@@ -699,8 +699,12 @@ pub const fn is_gregorian_valid(
     {
         return false;
     }
-    if day > usual_days_per_month(month) && (month != 2 || !is_leap_year(year)) {
-        // Not in February or not a leap year
+    let days_per_month = if month == 2 && is_leap_year(year) {
+        29
+    } else {
+        usual_days_per_month(month)
+    };
+    if day > days_per_month {
         return false;
     }
     true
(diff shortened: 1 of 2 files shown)


ephraim71:
@ChristopherRabotin One of the existing test cases `test_range()` is failing as `middle` is set to 30th Feb. Shall I change it to 29th Feb? 
```
#[test]
fn test_range() {
    let start = Epoch::from_gregorian_utc_hms(2012, 2, 7, 11, 22, 33);
    let middle = Epoch::from_gregorian_utc_hms(2012, 2, 30, 0, 11, 22);
    let end = Epoch::from_gregorian_utc_hms(2012, 3, 7, 11, 22, 33);
    let rng = start..end;
    assert_eq!(rng, core::ops::Range { start, end });
    assert!(rng.contains(&middle));
}
```

ChristopherRabotin:
Thanks for the fix!

Could you add the following test as a regression:

```
--- a/tests/epoch.rs
+++ b/tests/epoch.rs
@@ -2374,3 +2374,9 @@ fn test_et_continuity_around_j2000() {
         epoch_after_j2000_et - round_trip_after
     );
 }
+
+#[test]
+#[should_panic]
+fn test_regression_gh_440_february_epoch() {
+    let _ = Epoch::from_gregorian_tai(2024, 2, 31, 00, 00, 0, 0);
+}
```

Without your fix, this test fails, and with it, it passes.

Issue #422, opened 2025-09-26
Add `const` constructors for `Duration`


johan-aiko:
One advantage (or at least a difference) of `std::time::Duration` is that it can be constructed as a constant, which is not possible for `hifitime::Duration` as far as I can tell.

```rust
const MY_STD_DURATION: std::time::Duration = std::time::Duration::from_secs(5);
const MY_HIFI_DURATION: hifitime::Duration = hifitime::Duration::from_seconds(5.0); // <- Will not compile.
```

Would it be feasible to add public `const` constructors for `hifitime::Duration`?

ChristopherRabotin:
Hi Johan,

Duration types are [`normalize`d](https://github.com/nyx-space/hifitime/blob/master/src/duration/mod.rs#L319) when initialized so that negative durations correctly set the centuries counter to be negative, and the nanoseconds counter positive. The normalize function uses rem_euclid, div_euclid, and saturating_sub, neither of which are const (yet?)... which is why the Duration initializers cannot be const.

One option would be to create a set of initializers that ingest only unsigned integers. For example, I could create the `from_positive_seconds`, etc. that would mimic the [existing unit-based constructors](https://github.com/nyx-space/hifitime/blob/master/src/duration/mod.rs#L219-L252). Would that work for you? What function name do you propose, I'm not sure that `from_positive_X` is necessarily the best naming.

johan-aiko:
A couple of questions:

- Why does the input type for nanoseconds need to be less than a `u64`, and not less than or equal? Would `positive_from_nanoseconds(u64::MAX)` not result in a valid `Duration` (namely `Duration{centuries: 0, nanoseconds: u64::MAX})?

- Is an un-normalized duration considered valid? For example, if we construct `Duration{centuries:0, nanoseconds: NANOSECONDS_PER_CENTURY + 1}`, would this be a valid alternative to the normalized equivalent `Duration{centuries: 1, nanoseconds: 1}`?

Anyway, I will spend some time tinkering with this in the coming days, if I have time.

PR #423, opened 2025-09-29
Make unit constructors for `Duration` const-compatible

J-Westin:
The unit constructors for `Duration` (`from_nanoseconds`, `from_days`, etc.) are made `const`-compatible, allowing `Duration` to be declared as a compile-time constant. This is a fix for issue #422.

# Main Changes

- `Duration::normalize()` uses plain `/` and `%` operators in place of `div_euclid` and `rem_euclid`.

- A const-compatible `Duration::is_equal_to(other: Duration)` is added to replace two (non-)equality checks in `Duration::normalize()`.

- `Duration::from_total_nanoseconds()` is changed to use explicit `as i128` casts instead of `into()`.

- The existing implementation of `Mul<f64> for Unit` is moved to a non-trait `const` method called `Unit::const_multiply`. This method is called by the regular `mul`, and from the unit constructors.

# MSRV Compatibility

Additionally, for compatibility with the MSRV (1.82), the following changes have been made:

- The body of `Duration::normalize(&mut self)` is moved to a new constant method `Duration::to_normalized(&self)`. The existing `normalize(&mut self)` remains non-constant and instead calls `to_normalized()` on the input.

- A call to `f64::abs()` is replaced by a manual if/else implementation, since `f64::abs` was not `const`-compatible in the MSRV.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "hifitime"
-version = "4.2.0"
+version = "4.2.1"
 authors = ["Christopher Rabotin <christopher.rabotin@gmail.com>"]
 description = "Ultra-precise date and time handling in Rust for scientific applications with leap second support"
 homepage = "https://nyxspace.com/"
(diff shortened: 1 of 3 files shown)


ChristopherRabotin (commented):
Looks good, thanks! I recommend two small changes.

J-Westin:
@ChristopherRabotin I'm fine with having this released as  a patch version, I've bumped it in `Cargo.toml`.

ChristopherRabotin:
I'll release the new version soon, thanks for your contribution!

Issue #421, opened 2025-09-23
Conversion to Python's datetime of an Epoch in ephemeris time will be incorrect


ChristopherRabotin:
I expect the `todatetime` call to return the datetime object as `datetime.datetime(2025, 3, 7, 12, 1, 9, 185475)` but instead it seems that the ET reference epoch offset is not applied.

```
>>> e = Epoch("2025-03-07T12:01:09.185475585 ET")
>>> e
2025-03-07T12:00:00.000000009 UTC @ 0x7e61c3a48f70
>>> print(e)
2025-03-07T12:01:09.185475585 ET
>>> e.todatetime()
datetime.datetime(1925, 3, 8, 0, 1, 9, 185475)
>>> print(e.to_time_scale(TimeScale.UTC))
2025-03-07T12:00:00.000000009 UTC
>>> e.to_time_scale(TimeScale.UTC).todatetime()
datetime.datetime(2025, 3, 7, 12, 0)
```

J-Westin:
FYI, I've [added a test for this on my fork](https://github.com/J-Westin/jwestin-hifitime/commit/809522b9ac372abf49e516da019a0cdc8567b2b2), as a starting point.

PR #424, opened 2025-10-02
Fix conversion from Ephemeris Time to `datetime`

J-Westin:
A test is added recreating the test case given in #421. To make the test pass, a conversion of the input epoch's duration to UTC is added in `Epoch.todatetime()`.


ChristopherRabotin:
Superb, thanks!

PR #447, opened 2025-12-07
fix conflicting polynomial constructors

ChristopherRabotin:
Supersede #445


gwbres:
@ChristopherRabotin thank you

Issue #454, opened 2026-01-26
Missing Linux ARM64 wheels for Python 3.14


Jazzinghen:
Even though the release action seems to have built the wheels for all the target platforms (https://github.com/nyx-space/hifitime/actions/runs/19117556042), the Pypi page reports only a few of them:

<img width="735" height="260" alt="Image" src="https://github.com/user-attachments/assets/434cb1a5-a7c5-4d88-8cb2-91ef213b2170" />

Actually, it looks like the ARM64 wheels were built for all supported Python versions aside `3.14`:

<img width="662" height="458" alt="Image" src="https://github.com/user-attachments/assets/2a525fbe-9362-4063-ab80-5af70e231af4" />

Jazzinghen:
Ah, sorry.
I didn't see the message.
Yes, I can confirm that Hifitime is installed properly on ARM64 linux. Thank you for everything!

PR #455, opened 2026-01-28
Update CI pipeline 

Jazzinghen:
While exploring what happened with issue #454 I re-ran `maturin generate-ci --pytest -o .github/workflows/python.yml github` using `maturin v1.11.5`, which generated a new CI pipeline addressing a couple issues:

- Building everything on `x86_64` even though `aarch64` runners are available
- Deprecated `macos-13` runner in favor of `macos-15-intel`

I am not sure if this was the reason why ARM64 Python packages were not built for `4.2.3`. Based on the output of the Github CI workflow it seems like everything runs like before, however I miss part of the knowledge of how Hifitime is tested.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -30,7 +30,7 @@ jobs:
             target: x86
           - runner: ubuntu-22.04-arm
             target: aarch64
-          - runner: ubuntu-22.04-arm
+          - runner: ubuntu-22.04
             target: armv7
           - runner: ubuntu-22.04
             target: ppc64le
@@ -66,7 +66,7 @@ jobs:
           name: wheels-linux-${{ matrix.platform.target }}
           path: dist
       - name: pytest
-        if: ${{ startsWith(matrix.platform.target, 'x86_64') }}
+        if: ${{ matrix.platform.target == 'x86_64' || matrix.platform.target == 'aarch64' }}
         shell: bash
         run: |
           set -e
@@ -76,7 +76,7 @@ jobs:
           pip install pytest
           pytest
       - name: pytest
-        if: ${{ !startsWith(matrix.platform.target, 'x86') && matrix.platform.target != 'ppc64' }}
+        if: ${{ matrix.platform.target != 'x86_64' && matrix.platform.target != 'x86' && matrix.platform.target != 'aarch64'}}
         uses: uraimo/run-on-arch-action@v2
         with:
           arch: ${{ matrix.platform.target }}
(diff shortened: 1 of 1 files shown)


.github/workflows/python.yml
```diff
@@ -20,24 +20,32 @@ permissions:
...
-      
+          python-version: 3.x
       - name: Build wheels
         uses: PyO3/maturin-action@v1
         with:
-          target: ${{ matrix.target }}
-          args: --release --out dist --find-interpreter -F python
-          manylinux: ${{ matrix.target == 'aarch64' && 'manylinux_2_28' || 'auto' }}
+          target: ${{ matrix.platform.target }}
+          args: --release --out dist --find-interpreter
```
ChristopherRabotin:
```suggestion
          args: --release --out dist --find-interpreter -F python
```

I'm surprised that the wheels built successfully because I think the `-F python` flag is required to tell maturin to build with the `python` feature flag.

.github/workflows/python.yml
```diff
@@ -79,140 +88,183 @@ jobs:
...
-          architecture: ${{ matrix.target }}
+          python-version: 3.x
       - name: Build wheels
         uses: PyO3/maturin-action@v1
         with:
-          target: ${{ matrix.target }}
-          args: --release --out dist --find-interpreter -F python
-          sccache: 'true'
+          target: ${{ matrix.platform.target }}
+          args: --release --out dist --find-interpreter
```
ChristopherRabotin:
```suggestion
          args: --release --out dist --find-interpreter -F python
```

I expect that the feature flag also be needed here.

.github/workflows/python.yml
```diff
@@ -79,140 +88,183 @@ jobs:
...
+          architecture: ${{ matrix.platform.python_arch }}
       - name: Build wheels
         uses: PyO3/maturin-action@v1
         with:
-          target: x86_64
-          args: --release --out dist --find-interpreter -F python
-          sccache: 'true'
-
+          target: ${{ matrix.platform.target }}
+          args: --release --out dist --find-interpreter
```
ChristopherRabotin:
```suggestion
          args: --release --out dist --find-interpreter -F python
```

ChristopherRabotin (approved):
Thanks for the fix!

Now that I think more about the issue you had, I think I recently noticed a similar issue in ANISE and I think it's because the MacOS wheel packages have the same name for macos-14 and macos-15, despite the wheels being of different size. So it may be worth trying to rename the macos wheels.

Thanks for also changing to `uv` to publish: I recently had to implement that fix in ANISE as well.

PR #466, opened 2026-04-05
Fix 4 bugs found by Kani function contract proofs

feliperodri:
## Summary

This PR fixes four potential correctness bugs discovered by writing Kani proof harnesses that verify positive mathematical properties, not just panic-freedom. Each bug was invisible to the existing test suite (162 tests, all passing).

### Bug 1: `total_nanoseconds()` returns wrong value for negative durations

Severity: Silent data corruption for any Duration with centuries < -1.

The implementation had three branches: a special case for `centuries == -1`, one for `centuries >= 0`, and a fallback for `centuries < -1`. The fallback computed `centuries * NPC - nanoseconds`, but the internal representation stores nanoseconds counting toward zero, so the correct formula is `centuries * NPC + nanoseconds`. The fix replaces all three branches with the single universal formula `c * NPC + n`.

Contract: `#[kani::ensures]` specifies `result == c * NPC + n`. Verified by `proof_for_contract` over all i16 × u64 inputs (10s).

### Bug 2: `i64::MIN.abs()` panic in `Unit * i64`

Severity: Panic on `i64::MIN * Unit::Nanosecond`.

`checked_mul` succeeds returning `i64::MIN`, then `.abs()` panics because `i64::MIN` has no positive i64 representation. Fixed by replacing `abs()` with `unsigned_abs()`.

Proof: Verified for all i64 values including `i64::MIN` (23s). Cannot use `#[kani::ensures]` because Kani does not support contracts on generic trait methods ([#1997](https://github.com/model-checking/kani/issues/1997)).

### Bug 3: NaN and non-termination in Duration * f64 precision loop

Severity: Infinite loop when `q * 10^p` overflows to infinity.

The loop computes `q * 10^p` for increasing p until `floor(val) == val`. When the multiplication overflows to infinity, `floor(inf) - inf` produces NaN, and `NaN < EPSILON` is false, so the loop never breaks. Fixed by adding a `!is_finite()` guard and a `p >= 19` termination bound.

Contract: `#[kani::loop_invariant(p >= 0 && p <= 19)]` verified inductively (33s). Harness assumptions restrict the f64 range for CBMC tractability, they are verification budget constraints, not function preconditions (the function is correct for all f64 after the fix).

### Bug 4: `Epoch::PartialEq` and `Epoch::Ord` inconsistent at zero crossing

Severity: Violation of Rust's Ord contract i.e. two epochs can satisfy `a == b` and `a < b` simultaneously.

`Duration::PartialEq` deliberately treats opposite-sign durations as equal (-15min == +15min), which is valid for Duration as an abstract interval. But Epoch represents a point in time: +1ns and -1ns from J1900 are different points. `Epoch::PartialEq` inherited the sign-ignoring behavior while `Epoch::Ord` used the sign-aware derived `cmp`. Fixed by making both use `total_nanoseconds()`.

This bug is invisible to conventional testing because it only manifests for epochs constructed from raw Duration parts near the zero crossing, no Gregorian constructor ever produces such values. A regression test is included that demonstrates the non-obvious construction required to trigger it.

Proof: Verified over all i16 × u64 Duration values with zero assumptions (38s).

## Verification Results

| Proof | Property | Assumptions | Time |
|-------|----------|-------------|------|
| verify_total_nanoseconds_contract | result == c * NPC + n (proof_for_contract) | None | 10s |
| verify_mul_i64_no_panic | Normalized result, no panic | None | 23s |
| verify_mul_f64_terminates | Loop terminates, normalized result | f64 range (CBMC budget) | 33s |
| verify_epoch_eq_ord_consistent | eq(a,b) → cmp(a,b) == Equal | None | 38s |

### DIFF
--- a/src/duration/mod.rs
+++ b/src/duration/mod.rs
@@ -375,17 +375,13 @@ impl Duration {
 
     /// Returns the total nanoseconds in a signed 128 bit integer
     #[must_use]
+    #[cfg_attr(kani, kani::ensures(|result| {
+        *result == i128::from(self.centuries) * i128::from(NANOSECONDS_PER_CENTURY)
+                 + i128::from(self.nanoseconds)
+    }))]
     pub fn total_nanoseconds(&self) -> i128 {
-        if self.centuries == -1 {
-            -i128::from(NANOSECONDS_PER_CENTURY - self.nanoseconds)
-        } else if self.centuries >= 0 {
-            i128::from(self.centuries) * i128::from(NANOSECONDS_PER_CENTURY)
-                + i128::from(self.nanoseconds)
-        } else {
-            // Centuries negative by a decent amount
-            i128::from(self.centuries) * i128::from(NANOSECONDS_PER_CENTURY)
-                - i128::from(self.nanoseconds)
-        }
+        i128::from(self.centuries) * i128::from(NANOSECONDS_PER_CENTURY)
+            + i128::from(self.nanoseconds)
     }
 
     /// Returns the truncated nanoseconds in a signed 64 bit integer, if the duration fits.
--- a/src/epoch/ops.rs
+++ b/src/epoch/ops.rs
@@ -471,25 +471,12 @@ impl AddAssign<Duration> for Epoch {
 }
 
 /// Equality only checks the duration since J1900 match in TAI, because this is how all of the epochs are referenced.
-/// Equality compares epochs as points in time, using total nanoseconds for
-/// a canonical comparison that correctly distinguishes positive and negative
-/// durations (unlike Duration::PartialEq which treats opposite signs as equal).
+/// Equality compares epochs as points in time by normalizing to TAI.
+/// Uses field comparison via to_parts() to bypass Duration::PartialEq's
+/// zero-crossing special case (which treats opposite-sign durations as equal).
 impl PartialEq for Epoch {
     fn eq(&self, other: &Self) -> bool {
-        if self.time_scale == other.time_scale {
-            self.duration.total_nanoseconds() == other.duration.total_nanoseconds()
-        } else if self.time_scale.uses_leap_seconds() != other.time_scale.uses_leap_seconds() {
-            if self.time_scale.uses_leap_seconds() {
-                let converted = self.to_time_scale(other.time_scale).duration;
-                converted.total_nanoseconds() == other.duration.total_nanoseconds()
-            } else {
-                let converted = other.to_time_scale(self.time_scale).duration;
-                self.duration.total_nanoseconds() == converted.total_nanoseconds()
-            }
-        } else {
-            let converted = other.to_time_scale(self.time_scale).duration;
-            self.duration.total_nanoseconds() == converted.total_nanoseconds()
-        }
+        self.to_tai_duration().to_parts() == other.to_tai_duration().to_parts()
     }
 }
 
@@ -500,20 +487,16 @@ impl PartialOrd for Epoch {
 }
 
 impl Ord for Epoch {
-    /// Ordering uses total_nanoseconds() for consistency with PartialEq.
     fn cmp(&self, other: &Self) -> Ordering {
-        let self_dur = self.duration;
-        let other_dur = other.to_time_scale(self.time_scale).duration;
-        self_dur
-            .total_nanoseconds()
-            .cmp(&other_dur.total_nanoseconds())
+        self.to_tai_duration()
+            .to_parts()
+            .cmp(&other.to_tai_duration().to_parts())
     }
 }
 
 impl Hash for Epoch {
-    /// Hash normalizes to TAI nanoseconds so that two epochs representing the
-    /// same point in time produce the same hash, consistent with PartialEq.
+    /// Hash normalizes to TAI for consistency with PartialEq.
     fn hash<H: Hasher>(&self, state: &mut H) {
-        self.to_tai_duration().total_nanoseconds().hash(state);
+        self.to_tai_duration().hash(state);
     }
 }
--- a/src/epoch/ops.rs
+++ b/src/epoch/ops.rs
@@ -471,24 +471,24 @@ impl AddAssign<Duration> for Epoch {
 }
 
 /// Equality only checks the duration since J1900 match in TAI, because this is how all of the epochs are referenced.
+/// Equality compares epochs as points in time, using total nanoseconds for
+/// a canonical comparison that correctly distinguishes positive and negative
+/// durations (unlike Duration::PartialEq which treats opposite signs as equal).
 impl PartialEq for Epoch {
     fn eq(&self, other: &Self) -> bool {
         if self.time_scale == other.time_scale {
-            self.duration == other.duration
-        } else {
-            // If one of the two time scales does not include leap seconds,
-            // we always convert the time scale with leap seconds into the
-            // time scale that does NOT have leap seconds.
-            if self.time_scale.uses_leap_seconds() != other.time_scale.uses_leap_seconds() {
-                if self.time_scale.uses_leap_seconds() {
-                    self.to_time_scale(other.time_scale).duration == other.duration
-                } else {
-                    self.duration == other.to_time_scale(self.time_scale).duration
-                }
+            self.duration.total_nanoseconds() == other.duration.total_nanoseconds()
+        } else if self.time_scale.uses_leap_seconds() != other.time_scale.uses_leap_seconds() {
+            if self.time_scale.uses_leap_seconds() {
+                let converted = self.to_time_scale(other.time_scale).duration;
+                converted.total_nanoseconds() == other.duration.total_nanoseconds()
             } else {
-                // Otherwise it does not matter
-                self.duration == other.to_time_scale(self.time_scale).duration
+                let converted = other.to_time_scale(self.time_scale).duration;
+                self.duration.total_nanoseconds() == converted.total_nanoseconds()
             }
+        } else {
+            let converted = other.to_time_scale(self.time_scale).duration;
+            self.duration.total_nanoseconds() == converted.total_nanoseconds()
         }
     }
 }
@@ -500,15 +500,20 @@ impl PartialOrd for Epoch {
 }
 
 impl Ord for Epoch {
+    /// Ordering uses total_nanoseconds() for consistency with PartialEq.
     fn cmp(&self, other: &Self) -> Ordering {
-        self.duration
-            .cmp(&other.to_time_scale(self.time_scale).duration)
+        let self_dur = self.duration;
+        let other_dur = other.to_time_scale(self.time_scale).duration;
+        self_dur
+            .total_nanoseconds()
+            .cmp(&other_dur.total_nanoseconds())
     }
 }
 
 impl Hash for Epoch {
+    /// Hash normalizes to TAI nanoseconds so that two epochs representing the
+    /// same point in time produce the same hash, consistent with PartialEq.
     fn hash<H: Hasher>(&self, state: &mut H) {
-        self.duration.hash(state);
-        self.time_scale.hash(state);
+        self.to_tai_duration().total_nanoseconds().hash(state);
     }
 }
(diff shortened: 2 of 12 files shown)


src/epoch/ops.rs
```diff
@@ -471,24 +471,24 @@ impl AddAssign<Duration> for Epoch {
...
-            // If one of the two time scales does not include leap seconds,
-            // we always convert the time scale with leap seconds into the
-            // time scale that does NOT have leap seconds.
-            if self.time_scale.uses_leap_seconds() != other.time_scale.uses_leap_seconds() {
-                if self.time_scale.uses_leap_seconds() {
-                    self.to_time_scale(other.time_scale).duration == other.duration
-                } else {
-                    self.duration == other.to_time_scale(self.time_scale).duration
-                }
+            self.duration.total_nanoseconds() == other.duration.total_nanoseconds()
```
ChristopherRabotin: Why perform the check on total_nanoseconds? That forces a conversion to a single large integer instead of checking both parts of the Duration struct separately (less CPU cycles).

This comment applies to all other checks in this function.

src/epoch/ops.rs
```diff
@@ -500,15 +500,20 @@ impl PartialOrd for Epoch {
 }
 
 impl Ord for Epoch {
+    /// Ordering uses total_nanoseconds() for consistency with PartialEq.
     fn cmp(&self, other: &Self) -> Ordering {
-        self.duration
-            .cmp(&other.to_time_scale(self.time_scale).duration)
+        let self_dur = self.duration;
```
ChristopherRabotin: If there is an issue in `Ord` for Duration, then let's fix that instead of not using it in Epoch.

src/epoch/ops.rs
```diff
@@ -500,15 +500,20 @@ impl PartialOrd for Epoch {
...
     }
 }
 
 impl Hash for Epoch {
+    /// Hash normalizes to TAI nanoseconds so that two epochs representing the
+    /// same point in time produce the same hash, consistent with PartialEq.
     fn hash<H: Hasher>(&self, state: &mut H) {
-        self.duration.hash(state);
-        self.time_scale.hash(state);
+        self.to_tai_duration().total_nanoseconds().hash(state);
```
ChristopherRabotin: While we could switch everything to a monotonic time scale (like you've done here with TAI), maybe we could instead hash the time scale directly instead? I'm reluctant to use total_nanoseconds in general because it's a helper function and all Epoch calculations to should be offloaded to Duration after TimeScale conversions have happened.

src/timeunits.rs
```diff
@@ -255,7 +255,7 @@ impl Mul<i64> for Unit {
 
         match q.checked_mul(factor) {
             Some(total_ns) => {
-                if total_ns.abs() < i64::MAX {
+                if total_ns.unsigned_abs() < i64::MAX as u64 {
```
ChristopherRabotin, line 258: I'm not familiar with the difference between abs and unsigned_abs. Is it a different low-level implementation, i.e. is it more efficient than abs for example?

tests/epoch.rs
```diff
@@ -1206,6 +1206,50 @@ fn test_ord() {
...
+/// 3. The existing test_ord test above uses Gregorian dates (year 2020),
+///    which are far from the zero crossing.
+///
+/// The test below constructs epochs directly from Duration parts to hit
+/// the exact representation that triggers the inconsistency. Without
+/// formal verification, finding this would require knowing the internal
+/// Duration encoding AND the interaction between Duration::PartialEq's
+/// zero-crossing special case and the derived Duration::Ord.
+#[test]
+fn test_epoch_eq_ord_consistency_at_zero_crossing() {
```
ChristopherRabotin, line 1225: Great find! Thank you!

src/timeunits.rs
feliperodri: `unsigned_abs()` returns u64 instead of i64. The difference is correctness, not performance: `i64::MIN.abs()` panics (debug) or wraps to a negative value (release) because |i64::MIN| = 2^63 exceeds i64::MAX = 2^63 - 1. `i64::MIN.unsigned_abs()` returns 2^63 as u64, which can represent it. The comparison is updated to `< i64::MAX` as u64 to match the u64 return type. Our Kani proof `verify_mul_i64_no_panic` confirms this handles all i64 values including i64::MIN without panicking.

src/epoch/ops.rs
feliperodri: You're right! `total_nanoseconds()` forces an unnecessary i128 multiplication. AFAIU, the reason we can't use `self.duration == other.duration` is that `Duration::PartialEq` has a zero-crossing special case where (-1, NPC-1) equals (0, 1) (opposite-sign durations are treated as equal). For Epoch as a point in time, +1ns and -1ns from J1900 are different points, so we need to bypass that. But we don't need `total_nanoseconds()` to do it, since direct field comparison achieves the same correctness without the i128 cost:
```rust
// Instead of: self.duration.total_nanoseconds() == other.duration.total_nanoseconds()
// Use direct field comparison (bypasses Duration::PartialEq's zero-crossing logic):
self.duration.centuries == other.duration.centuries
    && self.duration.nanoseconds == other.duration.nanoseconds
```
This is O(1) with no multiplication, and it's correct because two Epochs at the same point in time with the same time scale MUST have identical (centuries, nanoseconds), thus, there's no reason for the zero-crossing equivalence when comparing points in time.

I'll update `PartialEq` and `Ord` to use field comparison for same-time-scale, and keep `total_nanoseconds()` only for the cross-time-scale path where we need a canonical scalar, **if you agree**.

src/epoch/ops.rs
feliperodri: Good point. My understanding is that the root issue is that `Duration::PartialEq` has a deliberate zero-crossing special case (-15min == 15min, documented in the struct doc as "opposite durations are equal"), but `Duration::Ord` is derived and doesn't have this, so they're already inconsistent at the Duration level.

The way I see, there are two paths:

1. Remove the zero-crossing from `Duration::PartialEq`. Make it a plain field comparison, consistent with derived `Ord`. This is the cleanest fix but it's a semantic change to Duration's documented behavior. Epoch can then just use `self.duration.cmp(&other.duration)` directly.

2. Keep the zero-crossing in Duration, but then `Duration::Ord` also needs to account for it (compare by absolute value), which changes sort order for negative durations.

Which do you prefer? I think I'm leaning towards option 1: the "opposite durations are equal" convention is surprising and the existing `Ord` already doesn't respect it, so removing it makes Duration internally consistent. I can update the PR to fix it at the Duration level instead.

src/epoch/ops.rs
feliperodri: The reason for normalizing to TAI in Hash is Rust's Hash/Eq contract: if a == b, then hash(a) == hash(b). Since `PartialEq` currently compares across time scales (e.g., a TAI epoch and a GPST epoch representing the same point in time are equal), hashing (`duration`, `time_scale`) directly would produce different hashes for equal epochs, silently breaking HashMap/HashSet.

If you'd prefer to hash (`duration`, `time_scale`) directly, we'd need to also change PartialEq so that epochs in different time scales are never equal (only same-time-scale comparison). IMHO, that's actually a much cleaner design. Users who want cross-time-scale comparison can explicitly convert first. Would you like me to go that direction instead?

That would make the whole `impl` simpler:

```rust
impl PartialEq for Epoch {
    fn eq(&self, other: &Self) -> bool {
        self.time_scale == other.time_scale
            && self.duration.centuries == other.duration.centuries
            && self.duration.nanoseconds == other.duration.nanoseconds
    }
}
impl Hash for Epoch {
    fn hash<H: Hasher>(&self, state: &mut H) {
        self.duration.hash(state);
        self.time_scale.hash(state);
    }
}
```

No `total_nanoseconds()`, no cross-time-scale conversion, all comparisons delegated to Duration after explicit time scale conversion by the caller.

src/epoch/ops.rs
ChristopherRabotin: If the timescales differ, why not simply convert one to the other's time scale like how it was originally done? I don't see the problem with the existing approach.

src/epoch/ops.rs
ChristopherRabotin: @gwbres , do you remember why `Duration::from_str("-15 min") == Duration::from_str("15 min")` ?? I just tested that, and it is indeed true ... but it doesn't make sense to me why we would ever want that to be the case. Was there a specific reason? It seems like a (major) bug.

src/epoch/ops.rs
ChristopherRabotin: > Users who want cross-time-scale comparison can explicitly convert first.

No, that's the whole use case of hifitime: it allows immediate comparison of epochs in different time scales: we use this functionality all the time in astrodynamics because, for example, trajectory files always store epochs in Ephemeris Time and we operate either in UTC or GPST.

To meet the HashMap contract, how about hashing the epoch in TAI every time, but skipping the total_nanoseconds? I tend to avoid using total_nanoseconds basically at all costs because it doubles the computation time since it requires two words on a 64 bit machine (over a year ago, I implemented a rewrite with i128 instead of the current structure, and noticed it was twice as slow, @gwbres explained to me why).

ChristopherRabotin (changes requested):

src/epoch/ops.rs
feliperodri: Hum... you got a point, we can use the original convert-then-compare approach. But we would still use direct field comparison instead of `Duration::PartialEq` after the conversion, to avoid the zero-crossing special case. Something like:
```rust
let converted = other.to_time_scale(self.time_scale);
self.duration.centuries == converted.duration.centuries
    && self.duration.nanoseconds == converted.duration.nanoseconds
```
Same logic as before, just bypassing `Duration::PartialEq`'s sign-ignoring behavior. No `total_nanoseconds()` anywhere.

src/epoch/ops.rs
feliperodri: I can do that! As soon as we agree on https://github.com/nyx-space/hifitime/pull/466#discussion_r3055420150, I can send an update 😄

src/epoch/ops.rs
ChristopherRabotin: Yes, that sounds good to me. Please also keep the logic with the branch based on whether only one of the two time scales includes leap seconds (leap seconds are a real pain and hifitime is the only rust time library that handle them (almost entirely correctly)).

feliperodri:
@ChristopherRabotin I implemented your comments, one last to go. I'm not too worried about the drop in coverage reported by codecov, but let me know if I also should update the tests.

src/duration/mod.rs
```diff
@@ -375,17 +375,13 @@ impl Duration {
...
-            -i128::from(NANOSECONDS_PER_CENTURY - self.nanoseconds)
-        } else if self.centuries >= 0 {
-            i128::from(self.centuries) * i128::from(NANOSECONDS_PER_CENTURY)
-                + i128::from(self.nanoseconds)
-        } else {
-            // Centuries negative by a decent amount
-            i128::from(self.centuries) * i128::from(NANOSECONDS_PER_CENTURY)
-                - i128::from(self.nanoseconds)
-        }
+        i128::from(self.centuries) * i128::from(NANOSECONDS_PER_CENTURY)
```
ChristopherRabotin, line 383: Your rewrite is simpler and I believe the original `else` clause was incorrect (when the centuries are negative by more than one).

src/epoch/ops.rs
```diff
@@ -471,24 +471,26 @@ impl AddAssign<Duration> for Epoch {
...
+/// Equality compares epochs as points in time.
+/// Uses field comparison via to_parts() to bypass Duration::PartialEq's
+/// zero-crossing special case (which treats opposite-sign durations as equal).
 impl PartialEq for Epoch {
     fn eq(&self, other: &Self) -> bool {
         if self.time_scale == other.time_scale {
-            self.duration == other.duration
-        } else {
-            // If one of the two time scales does not include leap seconds,
+            self.duration.to_parts() == other.duration.to_parts()
```
ChristopherRabotin, line 480: Yes! Thank you

ChristopherRabotin (approved):

PR #470, opened 2026-04-12
timeseries: derive Copy

gwbres:
Although it is made of copyable objects, the timeseries struct does not derive the Copy trait. Is there any reasons for that ?


ChristopherRabotin:
> Although it is made of copyable objects, the timeseries struct does not derive the Copy trait. Is there any reasons for that ?

Yes, the iterator holds its current progress in the iteration, so making it copy means you would copy not only the start and stop epoch but also how far it has progressed. Structs that implement copy are copied automatically when passed to functions: to prevent this possible hidden copy, timeseries implements clone and not copy.

PR #471, opened 2026-04-12
timeseries: derive Serdes

gwbres:
Although it is conveniently implemented for other structures, these traits are not available for the TimeSeries object

### DIFF
--- a/src/timeseries.rs
+++ b/src/timeseries.rs
@@ -38,7 +38,7 @@ NOTE: This is taken from itertools: https://docs.rs/itertools-num/0.1.3/src/iter
 /// :type inclusive: bool
 #[cfg_attr(kani, derive(kani::Arbitrary))]
 #[derive(Clone, Debug, PartialEq, Eq)]
-#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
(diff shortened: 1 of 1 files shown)


gwbres:
@ChristopherRabotin I see we use a manual implementation of the Serdes for Epoch, do you see any particular reasons we should do that for that object as well ?

ChristopherRabotin:
What is your use case for implement serde on time series? I have never had the need for it despite using it a lot : in some other code, I have helper functions that return the domain (start and stop), from which I initialize the time series.
My worry is similar to the copy : you probably don't want to serialize the cur and incl fields. It would be confusing if you serialized it after iterating through the time series. You'd always want to serialize it before iteration, in which case, you might as well just serialize the start, stop, and step without the cur and incl.

gwbres:
Hello, my use case is to describe a temporal arc, from a starting point and end point. But I'm starting to understand that, the answer is to create an object of my own, and convert it to Timeseries when I actually need to iterate. Timeseries cannot be used as the storage

ChristopherRabotin:
Yes, exactly, I advise against using the timeseries to store data. For example, anise can report event arcs (eg when the spacecraft is above the terrain of a location), and that is stored in its own structure which provides helper functions to compute the duration of the arc. https://docs.rs/anise/latest/anise/analysis/event/struct.EventArc.html

PR #473, opened 2026-04-13
Fix `Duration::PartialEq/Ord` inconsistency by removing zero-crossing special case

### LINKED ISSUE #469: Positive and negative durations are equal
Durations of opposite signs are considered equal. At the moment, I don't recall whether this was deliberate or whether this is a long-lived bug. I think one argument is that physically durations are absolute and therefore strictly positive. I need to dig into the history of version 4 to understand that decision.

_Originally posted by @ChristopherRabotin in https://github.com/nyx-space/hifitime/pull/466#discussion_r3055281339_

feliperodri:
Fixes #469

## Bug

`Duration::PartialEq` has a zero-crossing special case that treats opposite-sign durations as equal:

```rust
Duration::from_str("-15 min") == Duration::from_str("15 min") // returns true
```

However, `Duration::Ord` is derived (lexicographic comparison on (centuries, nanoseconds)) and does NOT have this special case. This means two Durations can satisfy a == b and a < b simultaneously, violating Rust's Eq/Ord contract.

This bug was discovered during formal verification of the Hifitime codebase with Kani. While proving `Epoch::PartialEq/Ord` consistency (PR #466), we traced the root cause to `Duration::PartialEq`'s zero-crossing logic.

## Fix

Remove the zero-crossing special case from `Duration::PartialEq`, making it a plain field comparison consistent with derived `Ord`:

```rust
impl PartialEq for Duration {
    fn eq(&self, other: &Self) -> bool {
        self.centuries == other.centuries && self.nanoseconds == other.nanoseconds
    }
}
```

## Impact

`-15.minutes() == 15.minutes()` now correctly returns false. Code that needs to compare absolute values should use `.abs()` explicitly:

```rust
// Before (buggy): -15.minutes() == 15.minutes()  → true
// After (correct): -15.minutes() == 15.minutes()  → false
// To compare magnitudes: (-15.minutes()).abs() == 15.minutes().abs()  → true
```

## Tests

Two existing tests relied on the buggy behavior and are updated:
- `test_neg`: `2.nanoseconds() == -(2.0.nanoseconds())` was asserting true; changed to `assert_ne!` and added double-negation identity test
- `epoch_stirring_static_offset`: compared a duration with its negation; changed to compare absolute values

New test `test_duration_eq_ord_consistency` verifies:
- Opposite-sign durations are not equal `(-15min != 15min)`
- Positive durations are greater than their negations
- `PartialEq` and `Ord` are consistent at the zero crossing

## Proof

Kani proof `verify_duration_eq_ord_consistent` formally verifies that `a == b` implies `a.cmp(&b) == Equal` for ALL possible Duration values (i16 × u64 — the full 2^80 input space). Verification time: 0.2s.

### DIFF
--- a/src/duration/mod.rs
+++ b/src/duration/mod.rs
@@ -59,7 +59,7 @@ pub mod ops;
 ///    Duration zero minus one nanosecond returns a century of -1 and a nanosecond set to the number of nanoseconds in one century minus one.
 ///    That difference is exactly 1 nanoseconds, where the former duration is "closer to zero" than the latter.
 ///    As such, the largest negative duration that can be represented sets the centuries to i16::MAX and its nanoseconds to NANOSECONDS_PER_CENTURY.
-/// 2. It was also decided that opposite durations are equal, e.g. -15 minutes == 15 minutes. If the direction of time matters, use the signum function.
+/// 2. Negative and positive durations are distinct: -15 minutes != 15 minutes. Use the signum function to check the sign, and abs() to get the absolute value.
 ///
 /// (Python documentation hints)
 /// :type string_repr: str
@@ -74,22 +74,7 @@ pub struct Duration {
 
 impl PartialEq for Duration {
     fn eq(&self, other: &Self) -> bool {
-        if self.centuries == other.centuries {
-            self.nanoseconds == other.nanoseconds
-        } else if (self.centuries.saturating_sub(other.centuries)).saturating_abs() == 1
-            && (self.centuries == 0 || other.centuries == 0)
-        {
-            // Special case where we're at the zero crossing
-            if self.centuries < 0 {
-                // Self is negative,
-                (NANOSECONDS_PER_CENTURY - self.nanoseconds) == other.nanoseconds
-            } else {
-                // Other is negative
-                (NANOSECONDS_PER_CENTURY - other.nanoseconds) == self.nanoseconds
-            }
-        } else {
-            false
-        }
+        self.centuries == other.centuries && self.nanoseconds == other.nanoseconds
     }
 }
 
--- a/tests/polynomial.rs
+++ b/tests/polynomial.rs
@@ -140,7 +140,7 @@ fn epoch_stirring_static_offset() {
         let dt = t_gpst_gst - t_ref_gpst;
 
         // What ever the time offset might be, the stirring should only produce an a0 offset
-        assert_eq!(dt, -(a0 * Unit::Nanosecond - offset_from_ref));
+        assert_eq!(dt.abs(), (a0 * Unit::Nanosecond - offset_from_ref).abs());
 
         // Linear operation: reciprocity applies at all times
         let reciprocal = t_gpst_gst
--- a/tests/polynomial.rs
+++ b/tests/polynomial.rs
@@ -140,7 +140,7 @@ fn epoch_stirring_static_offset() {
         let dt = t_gpst_gst - t_ref_gpst;
 
         // What ever the time offset might be, the stirring should only produce an a0 offset
-        assert_eq!(dt.abs(), (a0 * Unit::Nanosecond - offset_from_ref).abs());
+        assert_eq!(dt, a0 * Unit::Nanosecond - offset_from_ref);
 
         // Linear operation: reciprocity applies at all times
         let reciprocal = t_gpst_gst
--- a/src/duration/kani_verif.rs
+++ b/src/duration/kani_verif.rs
@@ -438,3 +438,20 @@ fn verify_mul_f64_terminates() {
             || (c == i16::MIN && n == 0)
     );
 }
+
+/// Proves Duration::PartialEq and Duration::Ord are consistent:
+/// if a == b then a.cmp(&b) == Equal, for all Duration values.
+/// This was Bug 5 (issue #469): the zero-crossing special case in
+/// PartialEq made -d == d, but derived Ord did not, violating
+/// Rust's Eq/Ord contract.
+#[kani::proof]
+fn verify_duration_eq_ord_consistent() {
+    let a: Duration = kani::any();
+    let b: Duration = kani::any();
+    if a == b {
+        assert!(
+            a.cmp(&b) == core::cmp::Ordering::Equal,
+            "PartialEq and Ord must be consistent"
+        );
+    }
+}
--- a/tests/duration.rs
+++ b/tests/duration.rs
@@ -236,11 +236,41 @@ fn test_ops_near_bounds() {
     assert_ne!(Duration::MAX - 1 * Unit::Nanosecond, Duration::MAX);
 }
 
+/// Regression test for https://github.com/nyx-space/hifitime/issues/469
+/// Duration::PartialEq previously treated opposite-sign durations as equal
+/// (-15min == 15min), which violated Rust's Eq/Ord contract since the derived
+/// Ord did not have this special case.
+#[test]
+fn test_duration_eq_ord_consistency() {
+    use hifitime::prelude::*;
+
+    // Opposite-sign durations must NOT be equal
+    assert_ne!(15.minutes(), -15.minutes());
+    assert_ne!(1.seconds(), -1.seconds());
+    assert_ne!(Duration::MIN_POSITIVE, Duration::MIN_NEGATIVE);
+
+    // Positive > Negative
+    assert!(15.minutes() > -15.minutes());
+    assert!(1.nanoseconds() > -1.nanoseconds());
+
+    // Eq and Ord must be consistent
+    let pos = Duration::from_parts(0, 1);
+    let neg = Duration::from_parts(-1, NANOSECONDS_PER_CENTURY - 1);
+    assert_ne!(pos, neg, "+1ns and -1ns must not be equal");
+    assert!(pos > neg, "+1ns must be greater than -1ns");
+
+    // Same-value durations are still equal
+    assert_eq!(15.minutes(), 15.minutes());
+    assert_eq!(-15.minutes(), -15.minutes());
+    assert_eq!(Duration::ZERO, Duration::ZERO);
+}
+
 #[test]
 fn test_neg() {
     assert_eq!(Duration::MIN_NEGATIVE, -Duration::MIN_POSITIVE);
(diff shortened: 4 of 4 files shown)


tests/polynomial.rs
```diff
@@ -140,7 +140,7 @@ fn epoch_stirring_static_offset() {
         let dt = t_gpst_gst - t_ref_gpst;
 
         // What ever the time offset might be, the stirring should only produce an a0 offset
-        assert_eq!(dt, -(a0 * Unit::Nanosecond - offset_from_ref));
+        assert_eq!(dt.abs(), (a0 * Unit::Nanosecond - offset_from_ref).abs());
```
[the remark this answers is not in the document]
feliperodri: That makes sense, `.abs()` weakens the test. Applied the suggestion: `dt` is exactly equal to `a0 * Unit::Nanosecond - offset_from_ref` (no negation needed). The original test used `-(...)` which happened to pass due to the buggy zero-crossing equality. The direct comparison is both correct and stronger.

src/duration/mod.rs
```diff
@@ -59,7 +59,7 @@ pub mod ops;
 ///    Duration zero minus one nanosecond returns a century of -1 and a nanosecond set to the number of nanoseconds in one century minus one.
 ///    That difference is exactly 1 nanoseconds, where the former duration is "closer to zero" than the latter.
 ///    As such, the largest negative duration that can be represented sets the centuries to i16::MAX and its nanoseconds to NANOSECONDS_PER_CENTURY.
-/// 2. It was also decided that opposite durations are equal, e.g. -15 minutes == 15 minutes. If the direction of time matters, use the signum function.
+/// 2. Negative and positive durations are distinct: -15 minutes != 15 minutes. Use the signum function to check the sign, and abs() to get the absolute value.
```
ChristopherRabotin, line 62: @gwbres I'd like your input on whether we should change this design. It does cause many test changes (to my surprise), but it would definitely trigger _at least_ a minor version change if we file this as a bug fix. There's an argument however that it is an architectural breaking change, requiring a new major version (I much prefer avoiding these).

gwbres: why not manually implement Ord as well ? would that solve the original issue ?

Yes, changing that implementation would be a major change I am afraid.

Negative durations do not physically make sense. Personnally I like the fundamental idea of following the laws of physics (but that's just my opinion and it might not be the best idea all of the time) . The only purpose of negative durations is to represent a point in time _before_ the "starting point" of a timescale. Maybe you could imagine a (not necessarily less painful) rework, where durations are limited to unsigned numbers, and the sign bit is introduced in the Epoch structure itself, since it does not serve any purpose elsewhere right ? But I don't see the benefit you'd get from that. And I don't foresee any convenient means to construct that epoch either  (writing simply  `-10minutes` is super convenient to the user)

ChristopherRabotin: @feliperodri , is there a way to manually implement Ord ? I imagine that a compatible Ord would simply compare these durations as unsigned, right?

feliperodri: I'm not sure what you mean by "manually implement Ord." The derived Ord does lexicographic comparison on `(centuries: i16, nanoseconds: u64)`. This happens to be correct for Duration's encoding:
1. Centuries comparison gives correct sign ordering: -2 < -1 < 0 < 1.
2. Within the same century, larger nanoseconds = closer to zero = larger value. For example, -5min is (-1, NPC - 300B) and -15min is (-1, NPC - 900B). Since NPC - 300B > NPC - 900B, derived Ord correctly says -5min > -15min.

The only problem was the custom PartialEq that treated opposite-sign durations as equal (-15min == 15min). The derived Ord never had this special case, so it was always correct, but it was inconsistent with `PartialEq`, violating Eq/Ord contract. The fix removes the zero-crossing special case from `PartialEq`, making it a plain field comparison, which is now consistent with the derived Ord. No changes to Ord at all IMHO.

I think that the "manually implement Ord to compare as unsigned" would actually be wrong. Treating the fields as unsigned would lose the sign information in centuries: i16. The current derived Ord (lexicographic on signed i16 then unsigned u64) seems like the correct total order for this encoding.

Now, versioning concern... this is a semantic change to `PartialEq` (previously -d == d, now -d != d), but the previous behavior violated Rust's own Eq/Ord contract. Code relying on -d == d was relying on unsound behavior that could cause silent corruption in `BTreeMap`, `sort()`, `dedup()`, etc. A minor version bump with a clear changelog entry seems appropriate to me. Calling it a bug fix rather than a breaking change, since the old behavior was a contract violation.

gwbres: > @feliperodri , is there a way to manually implement Ord ? I imagine that a compatible Ord would simply compare these durations as unsigned, right?

any trait can be implemented, or derived, you can decide to implement it yourself if the derived behavior is not "correct"

> The only problem was the custom PartialEq that treated opposite-sign durations as equal (-15min == 15min). The derived Ord never had this special case, so it was always correct, but it was inconsistent with PartialEq, violating Eq/Ord contract. [..] that could cause silent corruption in BTreeMap, sort(), dedup(), etc.

👍 ok I fully understand now

@ChristopherRabotinn why is PartialEq not derived for Durations, it seems pretty straightforward? 
for Epoch we can't auto derive because we need to reference to a common timescale, otherwise two identical points in time would be `!=` depending on the timescale they are referenced to. But for Durations, I don't understand the need to do so, other than creating the behavior explained in the current comments

ChristopherRabotin: The Duration struct does not derive PartialEq because then - 15 min and + 15 min would be different durations.

feliperodri: I think that's the core of the issue: the intent was for Duration to be sign-agnostic (treating -15min == +15min), but that intent was only implemented in `PartialEq`. The derived Ord is sign-aware (it compares centuries: i16 with its natural ordering), so -15min < +15min according to Ord but -15min == +15min according to `PartialEq`. Rust requires these to be consistent: `a == b` must imply `a.cmp(&b) == Equal`. There are two ways to fix this, IMHO:
1. Make Duration sign-aware everywhere (in this PR): remove the zero-crossing special case from `PartialEq`, keep the derived Ord. Now -15min != +15min and -15min < +15min. Very simple.
2. Make Duration sign-agnostic everywhere: keep the custom `PartialEq`, and also manually implement Ord to compare by absolute value (e.g., via `abs().to_parts()`). Then -15min == +15min and `-15min.cmp(&+15min) == Equal`. But this creates a different problem: we lose the ability to sort durations by their signed value, and `BTreeMap<Duration, _>` would merge -15min and +15min into the same key.

I think option 1 is simpler and more consistent with how the rest of the library works, i.e., arithmetic, display, and `signum()` are all sign-aware. The sign-agnostic equality was the exception, not the rule.

ChristopherRabotin: On version 4.2.6, two different epochs whose duration is exactly opposed are reported as equal. This is obviously a (major) bug we hadn't considered or tested for. Felipe fixed it in 040124f8 when changing the PartialEq to check the parts of the duration.

```rust
#[test]
fn test_neg_epoch_dur() {
    let dur = Unit::Minute * 15;

    let j2k_plus_15_min = Epoch::from_et_duration(dur);
    let j2k_minus_15_min = Epoch::from_et_duration(-dur);
    println!(
        "{j2k_minus_15_min}\t{j2k_plus_15_min} -> {}",
        j2k_minus_15_min == j2k_plus_15_min
    );
}
```

Output:

```
running 1 test
2000-01-01T11:45:00 ET	2000-01-01T12:15:00 ET -> true
test test_neg_epoch_dur ... ok
```

I agree with @gwbres that durations are physically always positive. I don't yet know how to reconcile the physics and this glaring bug.

gwbres: > I agree with @gwbres that durations are physically always positive. I don't yet know how to reconcile the physics and this glaring bug.

If I understand the issue shows up on the user side, when they store Epochs (in maps or treemaps) and they happen to have epochs before and after the startpoint of the timescale

feliperodri: The fix I proposed is actually consistent with physics in my perspective. In physics, a duration (time interval) is indeed always positive: it's a magnitude. But what Hifitime stores internally isn't a "duration" in the physics sense, it's a signed offset from a reference epoch. The Duration struct serves double duty:
1. As a time interval (always positive): "the meeting lasts 15 minutes"
2. As the internal storage of Epoch: a signed offset from the time scale's reference point (e.g., J1900 for TAI)                                                                                                                                   

For use case 1, the sign doesn't matter. Users construct durations with `15.minutes()` and they're always positive. For use case 2, the sign is essential. It's how Hifitime represents points in time before the reference epoch.

@ChristopherRabotin the bug you demonstrated is exactly the consequence: `Epoch::from_et_duration(dur)` and `Epoch::from_et_duration(-dur)` store Durations with opposite signs internally. The old PartialEq said those Durations are equal, so the Epochs compared as equal i.e. two points in time 30 minutes apart.                                                                                                                                                                                           

The fix (option 1 from the PR) makes Duration sign-aware in PartialEq, consistent with the derived Ord. Users who need the "physics duration" (unsigned magnitude) can use `.abs()`. This is the same convention as i64; it's signed, but you call `.abs()` when you only care about magnitude. The user-side impact @gwbres  mentions (Epochs in maps/treemaps) is the most dangerous consequence: any `BTreeMap<Epoch, _>` or `HashMap<Epoch, _>` where epochs straddle the reference point would silently merge distinct points in time into the same key.

gwbres:
@ChristopherRabotin,

I agree with a minor version change (4.x) as long as the API remains the same and we only improve internal behavior

feliperodri:
@ChristopherRabotin @gwbres just to recap, option 1 is already implemented on this branch, so the PR should be ready to merge as-is. The changes are:
  - `PartialEq`: zero-crossing special case removed, plain field comparison.
  - Ord: unchanged (derived, already correct).
  - Kani proof: `verify_duration_eq_ord_consistent` proves Eq/Ord consistency for all inputs.

No API changes, just corrected internal behavior.

src/duration/mod.rs
ChristopherRabotin: Thank you for your solid argumentation @feliperodri , I agree. I had to think about it for a little bit, hence the delay in my answer. Let's proceed.

PR #485, opened 2026-05-13
Fix Kani CI - matrix split, retry, and `dead_code` warning

feliperodri:
The weekly Kani verification workflow was failing because a single monolithic job running all 153 harnesses exceeded GitHub-hosted runner limits (7 GB RAM, 2 vCPUs). This PR splits the workflow into 11 parallel matrix jobs grouped by module and verification time, and fixes a `dead_code` warning on `stub_round`.

### 1. Fix `dead_code` warning on `stub_round` (`src/duration/kani_verif.rs`)
The `stub_round` function is only referenced via `#[kani::stub]` attributes, which the compiler doesn't see as a "use." Added `#[allow(dead_code)]`.

### 2. Split Kani CI into 11 matrix jobs (`.github/workflows/weekly-kani-checker.yaml`)

| Job | Harnesses | Est. time (sequential) |
|-----|-----------|----------------------|
| `epoch-contracts` | 48 | ~4 min |
| `epoch-proofs-light` | 24 | ~6 min |
| `epoch-eq-ord` | 1 | ~1.5 min |
| `epoch-proofs-heavy` | 10 | ~30 min |
| `duration-fast` | 32 | ~4 min |
| `duration-compose` | 2 | ~3.5 min |
| `duration-tz-offset` | 1 | ~2.5 min |
| `duration-floor` | 1 | ~5 min |
| `duration-subdivision` | 1 | ~6 min |
| `duration-decompose` | 1 | ~9 min |
| `other` | 22 | ~1 min |

All jobs use `-j 1` (sequential verification) to minimize peak memory. Parallelism comes from GitHub running the 11 jobs concurrently. `fail-fast: false` ensures one failure doesn't cancel the rest.

## Observed behavior on `ubuntu-latest`

In testing on my fork, **7 of 11 jobs pass consistently**. The remaining 4 (`epoch-proofs-light`, `duration-compose`, `duration-decompose`, `other`) fail intermittently due to runner preemption — not verification failures. The pattern is:
- All harnesses that complete before preemption show `VERIFICATION:- SUCCESSFUL`
- The job then exits with code 143 ("runner has received a shutdown signal")
- Different jobs fail on different runs (non-deterministic)

## Recommendation: larger runner

The free-tier `ubuntu-latest` runner (7 GB RAM, 2 vCPUs) is marginal for Kani-based verification. The SAT solver's memory footprint during the heaviest harnesses (`decompose`: 524s, `subdivision`: 350s) likely approaches the 7 GB limit, making the job vulnerable to preemption or OOM.

**If the project has access to GitHub Team/Enterprise runners**, switching to `ubuntu-latest-4core` (16 GB RAM, 4 vCPUs) would likely resolve all failures:

```yaml
runs-on: ubuntu-latest-4core  # 16 GB RAM, 4 vCPUs
```

This would also allow increasing `-j` back to 2–4 for faster wall-clock times. The matrix split remains valuable regardless of runner size for debuggability and resilience.

### DIFF
--- a/.github/workflows/weekly-kani-checker.yaml
+++ b/.github/workflows/weekly-kani-checker.yaml
@@ -7,8 +7,139 @@ on:
 
 jobs:
   kani-formal-model-checker:
-    name: Formal Verification
+    name: Kani (${{ matrix.module }})
     runs-on: ubuntu-latest
+    timeout-minutes: 120
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # Epoch proof_for_contract (48 harnesses, all <14s each)
+          - module: epoch-contracts
+            kani_args: >-
+              --harness kani_harness_Epoch_from_bdt
+              --harness kani_harness_Epoch_from_gpst
+              --harness kani_harness_Epoch_from_gst
+              --harness kani_harness_Epoch_from_qzsst
+              --harness kani_harness_Epoch_from_tai
+              --harness kani_harness_Epoch_from_tdb
+              --harness kani_harness_Epoch_from_tt
+              --harness kani_harness_Epoch_from_utc
+              --harness kani_harness_Epoch_from_et
+              --harness kani_harness_Epoch_from_jde
+              --harness kani_harness_Epoch_from_mjd
+              --harness kani_harness_Epoch_from_duration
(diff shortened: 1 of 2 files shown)


ChristopherRabotin:
I had totally missed this PR! Thanks for your contribution!

PR #491, opened 2026-06-22
Fix type-hint comment for Epoch.timedelta(), update stubs.

### LINKED ISSUE #489: Type stubs: time.Epoch.timedelta() should accept another Epoch
# Bug report

## Describe the bug
At https://github.com/nyx-space/anise/blob/master/anise-py/anise/time.pyi#L999, `Epoch.timedelta()` is type-hinted as:
```python
    def timedelta(self, other: Duration) -> Duration:
        """Differences between two epochs"""
```
Given the description, it seems like `other` should have type `Epoch`.

I'm happy to make a pull request, but wanted to make sure I'm right about this first.

cwastranis:
This fixes #489.

### DIFF
--- a/hifitime.pyi
+++ b/hifitime.pyi
@@ -2,7 +2,6 @@ from __future__ import annotations
 import datetime
 import typing
 
-
 @typing.final
 class Duration:
     """Defines generally usable durations for nanosecond precision valid for 32,768 centuries in either direction, and only on 80 bits / 10 octets.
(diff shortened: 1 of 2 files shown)


cwastranis:
I'm not sure where all the indentation changes in `hifitime.pyi` are coming from, is there a formatting step I am missing?

ChristopherRabotin:
Thanks!

The bindings are nominally generated using the generatr_stubs script, and all of the documented methods as in fact real methods in the rust binding. This ensures the pyi file maps to actual code.

cwastranis:
The changes to `hifitime.pyi` all come from running `.venv/bin/python3 generate_stubs.py hifitime hifitime.pyi`. This copies over the comment changes from `epoch/python.rs` but also creates a lot of whitespace differences.

Running yapf on the generated file (`.venv/bin/yapf -i hifitime.pyi`) leads to a different but still large set of whitespace changes.

ChristopherRabotin:
Thanks! There should be a --ruff option when running the stub generation, but it's okay, I can run that later.

cwastranis:
That did it, much better now. Looks like it did generate some incorrect stubs for `__setattr__()` and `__delattr__()` that weren't there previously. I'm not sure why that would be - generate_stubs looks like it has correct `BUILTINS` entries for these functions: https://github.com/nyx-space/hifitime/blob/master/generate_stubs.py#L53

Let me know if you'd like me to look into this further.

ChristopherRabotin:
It seems like these new magic methods might be available due to a change in pyo3? I don't think it's a big deal. I'll merge that asap, thanks!

PR #495, opened 2026-07-29
Handle %w parsing and weekday-only formatting instead of panicking

gaoflow:
The `efmt` strftime module has two code paths that panic on valid, in-domain input reachable from the safe public API (both shipped in 4.3.0):

**1. Parsing `%w` panics with `todo!()`** (`src/efmt/format.rs`, `Format::parse`):

```rust
let e = Epoch::from_format_str("3 2026-07-29", "%w %Y-%m-%d");
// thread panicked: not yet implemented
```

**2. Formatting a weekday-only format panics with `unreachable!()`** (`src/efmt/formatter.rs`):

```rust
let e = Epoch::from_gregorian_utc_at_midnight(2026, 7, 29);
let _ = Formatter::new(e, Format::from_str("%A").unwrap()).to_string();
// thread panicked: internal error: entered unreachable code
```

Both share one theme: the weekday tokens were only partially wired up. `%A`/`%a` format fine when combined with a Gregorian token (e.g. `%A %Y-%m-%d`) because `need_gregorian()` then routes through the branch that implements them, but on their own they fall into the reduced branch's `_ => unreachable!()`. And the formatter emits a `%w` value that `parse` could not read back, so a `%w` round-trip crashed.

Fixes:

- `%w` parsing now mirrors the existing numeric-token arm: parse the decimal, run the already-present `WeekdayDecimal` `value_ok` range check (0..=6, previously unreachable), and invert `to_c89_weekday` (C89 `Sunday=0..Saturday=6`) back to the internal `Monday=0..Sunday=6` ordering. The parsed weekday feeds the existing consistency check, so a `%w` that contradicts the date is rejected with `WeekdayMismatch`, exactly like the `%A`/`%a` name tokens. It validates rather than silently accepts.
- `need_gregorian()` now returns true for `Weekday`/`WeekdayShort`, routing weekday-only formats through the branch that already handles them. Any format affected by this previously hit the `unreachable!()`, so there is no prior working behaviour to change.

I audited the rest of `efmt` for the same shape: these are the only two panic sites reachable from valid public input. The remaining `todo!()`/`unreachable!()` arms in `format.rs` are dead (those tokens are consumed by explicit arms earlier in the match), so I left them untouched to keep the diff minimal.

Tests added to `tests/efmt.rs` cover the full inverse map over a week, `%w` format round-trips, the date-mismatch rejection, and out-of-range rejection. Verified `cargo test`, `--no-default-features`, and `--features serde` all pass, plus `cargo fmt`. The change is confined to `efmt`, which is not under the Kani proofs, so no proof is affected.

### DIFF
--- a/src/efmt/format.rs
+++ b/src/efmt/format.rs
@@ -135,12 +135,15 @@ impl Format {
                     | Token::Second
                     | Token::Subsecond
                     | Token::OffsetHours
-                    | Token::OffsetMinutes => return true,
+                    | Token::OffsetMinutes
+                    // The Weekday name tokens need the Gregorian date so the reduced
+                    // (non-Gregorian) formatting branch does not have to handle them;
+                    // that branch only implements WeekdayDecimal.
+                    | Token::Weekday
+                    | Token::WeekdayShort => return true,
                     Token::Timescale
                     | Token::DayOfYearInteger
                     | Token::DayOfYear
-                    | Token::Weekday
-                    | Token::WeekdayShort
                     | Token::WeekdayDecimal => {
                         // These tokens don't need the gregorian, but other tokens in the list of tokens might.
                     }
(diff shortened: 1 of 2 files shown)


ChristopherRabotin (approved):
Super, merci beaucoup! I guess I never implemented these code paths because I don't typically use the weekday formatting, so thank you for fixing this bug.

When do you need this by? I can make a new release this weekend, or earlier if you need it right away.

ChristopherRabotin:
The test failure is due to the JPL servers, not your change.

Issue #186, opened 2022-12-03
Increase precisions from 10^-29 seconds to 10^28 seconds


ChristopherRabotin:
This is really just for fun and I don't imagine anyone ever needing this, but behind an `ultra` feature crate, the `Duration` struct could change the `centuries` from `i16` to `i64` and add a `subnanos` field of type `u64`. This would make the structure relatively big (192 bits / 24 octets) but it would allow representations of duration from 5.4e-29 seconds (or 54 quectoseconds) to 2.9e+28 seconds (2.9e+4 yottaseconds or 29 qettaseconds).

According to https://en.wikipedia.org/wiki/Orders_of_magnitude_(time), this means that the smallest duration would be 162 thousand times _longer_ than the [mean lifetime](https://en.wikipedia.org/wiki/Mean_lifetime) of [W and Z bosons](https://en.wikipedia.org/wiki/W_and_Z_bosons) and the largest time would be _just short_ of the radioactive half-life of [tellurium-128](https://en.wikipedia.org/wiki/Tellurium-128), the longest known half-life of any elemental [isotope](https://en.wikipedia.org/wiki/Isotope).

All in one library! \o/

Math:
```python
In [8]: 1.0/2**64 * 1e-9 # they are sub-nanoseconds hence the 1e-9
Out[8]: 5.421010862427523e-29
(...)
In [14]: (Unit.Century * 1).to_seconds()
Out[14]: 3155760000.0

In [15]: 2**(64-1) * 3155760000.0 # we can't do the math with Durations since they max out
Out[15]: 2.9106748539024827e+28
```

ChristopherRabotin:
More seriously, it might make sense to have a crate feature called `attosecond` where simply a `u32` would be added as subnanos. 

Adding a `u32` as `subnanos` would support 0.2328... attoseconds. Note that to support all of the specs of the Precise Time Protocol, femtoseconds may be needed since the maximum allowable Jitter is 250 fm: https://github.com/opencomputeproject/Time-Appliance-Project/tree/master/Open-Time-Server/#pps-out .

```python
In [3]: 1.0/2**32 * 1e-9
Out[3]: 2.3283064365386964e-19
```

ChristopherRabotin:
I should run some benchmarks to see if there is any impact in changing the centuries to a 32 bit integer and adding that subnanos field. The fields of `Duration` are not exposed publicly, so changing those won't be a breaking change.

ChristopherRabotin:
[...]

PR #496, opened 2026-07-30
Fix Duration subtraction saturation at bounds

### LINKED ISSUE #494: `Duration::ZERO - Duration::MIN` and `-Duration::MIN` disagree
```rust
use hifitime::Duration;

fn main() {
    println!("Duration::ZERO - Duration::MIN = {:?}", Duration::ZERO - Duration::MIN);
    println!("-Duration::MIN               = {:?}", -Duration::MIN);
    println!("Duration::MIN                = {:?}", Duration::MIN);
    println!("Duration::MAX                = {:?}", Duration::MAX);
}
```

```
Duration::ZERO - Duration::MIN = Duration { centuries: -32768, nanoseconds: 0 }
-Duration::MIN               = Duration { centuries: 32767, nanoseconds: 3155760000000000000 }
Duration::MIN                = Duration { centuries: -32768, nanoseconds: 0 }
Duration::MAX                = Duration { centuries: 32767, nanoseconds: 3155760000000000000 }
```

`Duration::ZERO - Duration::MIN` is mathematically the same operation as `-Duration::MIN`, and unary negation correctly saturates it to `Duration::MAX`. But going through `Sub` instead returns `Duration::MIN` unchanged, as if the subtraction had no effect at all.

Tested on hifitime 4.3.0.

BTW, this bug was found using [hegel](https://crates.io/crates/hegeltest). Happy to contribute the tests if you're interested.

LunaMeerkats:
## Summary

- compute `Duration` subtraction in full-width nanoseconds before clamping the result
- preserve representable values when a nanosecond borrow crosses the century boundary
- cover exact bounds, saturation in both directions, and both adjacent representable values

The regression test failed before the fix because `Duration::ZERO - Duration::MIN` returned `Duration::MIN` instead of `Duration::MAX`.

Fixes #494

## Validation

- `cargo fmt --all -- --check`
- `cargo check`
- `cargo clippy -- -D warnings`
- `cargo test`
- `cargo test --features ut1`
- `cargo test --features lts`
- `cargo test --no-default-features`
- `cargo test --features serde`

Kani, the MSRV 1.85 matrix, and bare-metal nightly targets were not available on this Windows host and are left to the repository workflows.

## Provenance

This patch was authored and validated by OpenAI Codex operating autonomously on behalf of @LunaMeerkats.

### DIFF
--- a/src/duration/ops.rs
+++ b/src/duration/ops.rs
@@ -302,35 +302,7 @@ impl Sub for Duration {
         // Ensure that the durations are normalized to avoid extra logic to handle under/overflows
         self.normalize();
         rhs.normalize();
-        match self.centuries.checked_sub(rhs.centuries) {
-            None => {
-                // Underflowed, so we've hit the min
-                return Self::MIN;
-            }
-            Some(centuries) => {
-                self.centuries = centuries;
-            }
-        }
-
-        match self.nanoseconds.checked_sub(rhs.nanoseconds) {
-            None => {
-                // Decrease the number of centuries, and realign
-                match self.centuries.checked_sub(1) {
-                    Some(centuries) => {
-                        self.centuries = centuries;
-                        self.nanoseconds += NANOSECONDS_PER_CENTURY - rhs.nanoseconds;
-                    }
-                    None => {
-                        // We're at the min number of centuries already, and we have extra nanos, so we're saturated the duration limit
-                        return Self::MIN;
-                    }
-                };
-            }
-            Some(nanos) => self.nanoseconds = nanos,
-        };
-
-        self.normalize();
-        self
+        Self::from_total_nanoseconds(self.total_nanoseconds() - rhs.total_nanoseconds())
     }
 }
 
--- a/src/duration/ops.rs
+++ b/src/duration/ops.rs
@@ -302,7 +302,35 @@ impl Sub for Duration {
         // Ensure that the durations are normalized to avoid extra logic to handle under/overflows
         self.normalize();
         rhs.normalize();
-        Self::from_total_nanoseconds(self.total_nanoseconds() - rhs.total_nanoseconds())
+
+        // Widen the century subtraction so that its direction is still known
+        // when the result exceeds the Duration bounds.
+        let mut centuries = i32::from(self.centuries) - i32::from(rhs.centuries);
+        let mut nanoseconds = match self.nanoseconds.checked_sub(rhs.nanoseconds) {
+            Some(nanoseconds) => nanoseconds,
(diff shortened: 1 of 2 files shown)


ChristopherRabotin:
Thanks for your PR.

I'll need to review this, and the ticket description, more closely. This whole block isn't accidental : it avoids switching to an i128 for calculations while still being correct.

For version 5, we have considered switching the whole code to i128 (cf #186) but that leads to a significant performance hit as all operations require two 64 bit registers.

LunaMeerkats:
Thanks @ChristopherRabotin — you were right about the `i128` cost. I measured it before revising the patch: on Windows 11 with rustc 1.96.1, a release-mode 50-million-iteration `black_box(lhs - rhs)` probe produced these five-run medians:

- base `7b6238d`: 2.401 ns/op (2.177–2.922)
- prior `i128` patch: 12.046 ns/op (11.720–12.617)
- revised `e8a3e9f`: 2.353 ns/op (2.325–2.424)

The new commit widens only the century difference to `i32`, handles the `u64` nanosecond borrow/carry explicitly, and clamps after the widened result is known. It performs no `i128` conversion, multiplication, division, or remainder. I also added a non-saturating `MAX` carry regression.

The focused regression, all 20 duration tests, default/UT1/LTS/no-default/serde suites, no-default check and Clippy with warnings denied, formatting, and diff checks pass locally. The four new workflow runs are approval-gated again; I’ll inspect every job after they execute.

src/duration/ops.rs
```diff
@@ -302,35 +302,35 @@ impl Sub for Duration {
...
-                        self.centuries = centuries;
-                        self.nanoseconds += NANOSECONDS_PER_CENTURY - rhs.nanoseconds;
-                    }
-                    None => {
-                        // We're at the min number of centuries already, and we have extra nanos, so we're saturated the duration limit
-                        return Self::MIN;
-                    }
-                };
+                centuries -= 1;
+                NANOSECONDS_PER_CENTURY - (rhs.nanoseconds - self.nanoseconds)
```
ChristopherRabotin, line 313: This line effectively does the normalization, which is good, because it means the `normalize` call does not need to do any work. And in fact, it speeds up the duration benchmarks by quite a bit!

Previously: https://github.com/nyx-space/hifitime/actions/runs/30633091528/attempts/1#summary-91163904748

Notably the "Convert f64 duration to hifitime duration" now takes 2.8 ns instead of 4.5 ns (consistent across previous benchmark runs).

ChristopherRabotin (approved):
Nice work, thank you

PR #497, opened 2026-07-31
Fix {:o} formatting panic outside the GPS u64 range

gaoflow:
`{:o}` on an Epoch panics for any instant outside the u64 GPST-nanosecond window (before 1980-01-06 or after ~2080): `to_gpst_nanoseconds()` returns an overflow error there and the Octal impl unwraps it. Formatting 1970-01-01 UTC, the UNIX epoch, panics. Our earlier #495 fixed the two panic sites in `efmt`; this one is outside `efmt`, in the `fmt::Octal` impl (formatting.rs:131), which that audit missed.

The fix formats the signed total GPS nanoseconds (`Duration::total_nanoseconds()`, which never overflows) instead of the fallible u64. In-range output is unchanged (the other format traits in this file use the letter as a time-scale access key — `{:x}` is TAI, `{:p}` is unix — so the decimal GPST-nanosecond output stands). Pre-1980 epochs now print the negative offset (1970-01-01 -> `-315964819000000000`), and post-2080 epochs no longer panic. TimeSeries `{:o}` is covered as well, since it formats both bounds through the Epoch impl.

Added boundary tests either side of the GPS epoch and a TimeSeries regression; `cargo test` on all CI feature variants, clippy, and fmt are clean.

### DIFF
--- a/src/epoch/formatting.rs
+++ b/src/epoch/formatting.rs
@@ -128,6 +128,6 @@ impl fmt::Pointer for Epoch {
 impl fmt::Octal for Epoch {
     /// Prints the Epoch in GPS
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}", self.to_gpst_nanoseconds().unwrap())
+        write!(f, "{}", self.to_gpst_duration().total_nanoseconds())
     }
 }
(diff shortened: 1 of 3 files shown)


ChristopherRabotin (approved):
Looks good, thank you!

PR #498, opened 2026-08-05
Switch latest leap seconds to static instead of const

ChristopherRabotin:
In theory this should lead to a speedup, let's check the benchmarks.

Results: 
- TBD seconds and JDE ET: sped up by 100 ns
- TT: sped up by 50 ns

Added a new benchmark that converts form UTC to ET.

On this branch:

```
UTC to ET               time:   [486.73 ns 492.90 ns 499.95 ns]
Found 9 outliers among 100 measurements (9.00%)
  3 (3.00%) high mild
  6 (6.00%) high severe
```

On master (after running on this branch):

```
TBD seconds and JDE ET  time:   [1.1241 µs 1.1321 µs 1.1416 µs]
                        change: [+1.4889% +2.9622% +4.2750%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild

UTC to ET               time:   [488.10 ns 491.21 ns 494.29 ns]
                        change: [+1.3694% +2.2958% +3.3155%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe

TT                      time:   [409.03 ns 411.95 ns 415.02 ns]
                        change: [−0.3812% +0.6389% +1.5179%] (p = 0.21 > 0.05)
                        No change in performance detected.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild
```

### DIFF
--- a/src/epoch/leap_seconds.rs
+++ b/src/epoch/leap_seconds.rs
@@ -44,7 +44,7 @@ impl LeapSecond {
         }
     }
 }
-
+#[cfg(not(kani))]
 static LATEST_LEAP_SECONDS: [LeapSecond; 42] = [
     LeapSecond::new(1_893_369_600.0, 1.417818, false), // SOFA: 01 Jan 1960
     LeapSecond::new(1_924_992_000.0, 1.422818, false), // SOFA: 01 Jan 1961
@@ -93,8 +93,8 @@ static LATEST_LEAP_SECONDS: [LeapSecond; 42] = [
 /// List of leap seconds from <https://data.iana.org/time-zones/data/leap-seconds.list>.
 /// This list corresponds the number of seconds in TAI to the UTC offset and to whether it was an announced leap second or not.
 /// The unannoucned leap seconds come from dat.c in the SOFA library.
+#[cfg(not(kani))]
 #[cfg_attr(feature = "python", pyclass)]
-#[cfg_attr(kani, derive(kani::Arbitrary))]
 #[derive(Clone, Debug)]
 pub struct LatestLeapSeconds {
     data: &'static [LeapSecond; 42],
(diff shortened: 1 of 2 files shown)


ChristopherRabotin:
@feliperodri , do you know if a static reference is supported by Kani arbitrary? Maybe I need to manually derive Arbitrary instead? This PR very slightly improves the performance of time scale conversions when leap seconds are involved.

feliperodri:
> @feliperodri , do you know if a static reference is supported by Kani arbitrary? Maybe I need to manually derive Arbitrary instead? This PR very slightly improves the performance of time scale conversions when leap seconds are involved.

Sorry for the late reply @ChristopherRabotin, but **no, Kani does not implement `Arbitrary` for reference types.** Upstream `kani_core/src/arbitrary.rs` has impls for `[T; N]`, tuples, `Option`/`Result`, `Range*`, `MaybeUninit`, `PhantomData`... but nothing for `&T`. So `derive(kani::Arbitrary)` can never work on a struct holding `&'static [LeapSecond; 42]`.

But I don't think we should work around it, because it isn't a limitation to route around, this is deliberate. There's an upstream PR from last week, [model-checking/kani#4694](https://github.com/model-checking/kani/pull/4694), that makes Kani *explicitly refuse* to derive `Arbitrary` for structs with reference fields. The reason is worth quoting, because it means the compile error is doing us a favour:

> The compiler-synthesized `any()` then created the referent's storage inside its own body and returned a dangling reference, so the generated harness failed spuriously with "dereference failure: dead object" — a false alarm the user can do nothing about.

So if `derive` had "worked" here, we'd have traded a compile error for a bogus verification failure.

src/epoch/leap_seconds.rs
```diff
@@ -149,6 +149,73 @@ impl LatestLeapSeconds {
     }
 }
 
+#[cfg(not(kani))]
```
feliperodri: @ChristopherRabotin I'd rather not merge the `#[cfg(kani)]` duplicate, for two reasons.

**1. It duplicates the 42-entry leap-second table.** That table is the crate's source of truth for IERS leap seconds. With two copies, the next announced leap second has to be added twice, and nothing will catch it if only one is updated and verification would then quietly prove properties about a stale table. That's a bad failure mode for a leap-second library specifically.

**2. It isn't buying us anything.** I grepped every harness on this branch — `src/kani_verif.rs`, `src/duration/kani_verif.rs`, `src/timescale/kani.rs`, `src/efmt/kani_verif.rs`, `src/epoch/kani_verif.rs`, `src/epoch/gregorian.rs` — and nothing ever constructs `kani::any::<LatestLeapSeconds>()`. The only use of the type anywhere in the crate is `LatestLeapSeconds::default()` at `src/epoch/mod.rs:439`. And the leap-second paths are already excluded from verification on purpose: `Epoch::leap_seconds` is handled with `#[kani::stub(...)]`, and there's an explicit comment that "Kani struggles with the leap second counting logic."

In other words that `derive` is dead weight. Deleting it keeps the `static` perf win, keeps one table, and removes all the `cfg` forking:

```suggestion
#[cfg_attr(feature = "python", pyclass)]
#[derive(Clone, Debug)]
pub struct LatestLeapSeconds {
    data: &'static [LeapSecond; 42],
    iter_pos: usize,
}
```

…and then the whole `#[cfg(kani)]` block at the end (the duplicated `LATEST_LEAP_SECONDS` const, the second `LatestLeapSeconds` struct, and its `Default` impl) can go away, along with the `#[cfg(not(kani))]` attributes on the `static`, the struct, and its `Default`.

Two notes:

- **Keep** `#[cfg_attr(kani, derive(kani::Arbitrary))]` on `LeapSecond` itself. That one is a plain POD struct, it derives fine, and `kani_harness_LeapSecond_new` actually uses it.

If all of this has some issue I didn't foreseen, the right shape might be a small hand-written impl that borrows the single real table and only makes the cursor symbolic:

```rust
#[cfg(kani)]
impl kani::Arbitrary for LatestLeapSeconds {
    fn any() -> Self {
        Self {
            data: &LATEST_LEAP_SECONDS,
            iter_pos: kani::any(),
        }
    }
}
```

That gives symbolic iteration over the real data with no duplication, but I'd only add it once a harness needs it.

ChristopherRabotin: Thanks for jumping in. I agree with both of your arguments. I'll implement them right away.

PR #508, opened 2026-09-04
Exact Duration scaling by f64

### LINKED ISSUE #502: Duration / f64 returns incorrect results for non-integral divisors
`Duration / f64` truncates the divisor to whole nanoseconds before dividing, so any
non-integral divisor gives a wrong answer.

```rust
use hifitime::TimeUnits;

assert_eq!(1.hours() / 2.5, 24.minutes()); // fails: returns 30 min
assert_eq!(1.hours() / 1.5, 40.minutes()); // fails: returns 1 h
```

| expression | actual | expected |
|---|---|---|
| `1.hours() / 2.5` | 30 min | 24 min |
| `1.hours() / 1.5` | 1 h | 40 min |

Integral divisors (`/ 2.0`, `/ 3.0`) are correct, which is why this hides well.

The same truncation makes any `0 < abs(q) < 1` divisor collapse to a zero divisor
and panic:

```rust
let _ = 1.hours() / 0.5;  // panicked at 'attempt to divide by zero'
let _ = 1.hours() / 0.1;  // panicked at 'attempt to divide by zero'
```

Root cause: `impl Div<f64> for Duration` divides by
`(q * Unit::Nanosecond).total_nanoseconds()`, which is `q as i64` nanoseconds.

Version: 4.3.1

### LINKED ISSUE #503: Duration / 0 panics instead of saturating
Every other `Duration` operator saturates at `Duration::MAX` / `Duration::MIN`
rather than panicking (`Add`, `Sub`), but division by zero traps.

```rust
use hifitime::TimeUnits;

let _ = 1.hours() / 0i64;  // panicked at 'attempt to divide by zero'
let _ = 1.hours() / 0.0;   // panicked at 'attempt to divide by zero'
```

Expected, to match the saturating behaviour of the other operators:

| expression | expected |
|---|---|
| `1.hours() / 0` | `Duration::MAX` |
| `(-1).hours() / 0` | `Duration::MIN` |
| `Duration::ZERO / 0` | `Duration::ZERO` |

Version: 4.3.1

### LINKED ISSUE #504: Duration * f64::NAN returns a value determined by the NaN's sign bit
Scaling by NaN saturates, and which bound you get depends only on the sign bit of
the NaN -- a bit that carries no meaning. The duration's own sign is ignored.

```rust
use hifitime::{Duration, TimeUnits};

dbg!(1.hours()    * f64::NAN);   // Duration::MAX
dbg!((-1).hours() * f64::NAN);   // Duration::MAX  <- duration sign ignored
dbg!(1.hours()    * -f64::NAN);  // Duration::MIN  <- only the NaN's sign bit decides
```

NaN sign bits are not preserved consistently across float operations, so this makes
the result of a poisoned computation effectively arbitrary. Whatever the right answer
is, it shouldn't be selected by that bit.

Version: 4.3.1

### LINKED ISSUE #505: No way to detect a non-finite f64 scale factor, and the behaviour is undocumented
There is no documented or recoverable answer to "what happens if my scale factor is
NaN or infinite?" The library has two different policies and says so nowhere:

| API | non-finite input |
|---|---|
| `Duration * f64`, `Duration / f64` | saturates |
| `Freq * f64` | returns `Duration::ZERO` for NaN |
| `Duration::from_seconds` and friends, `Unit * f64` | panics |
| `Epoch + f64` | panics |
| `Polynomial::from((f64, f64, f64))` | panics |

Two requests:

1. Document the split. The panicking constructors have no `# Panics` sections.
2. Add a fallible scaling API. `std` pairs `Duration::from_secs_f64` (panics) with
   `try_from_secs_f64` (returns `Result`); hifitime has no equivalent, so the only
   way to handle a possibly-NaN factor is an `is_finite()` check at every call site.

Version: 4.3.1

### LINKED ISSUE #506: Duration is missing MulAssign / DivAssign
`AddAssign` and `SubAssign` are implemented, but the multiplicative equivalents are not.

```rust
use hifitime::TimeUnits;

let mut d = 1.hours();
d *= 2.0;  // error[E0368]: binary assignment operation `*=` cannot be applied
           //               to type `hifitime::Duration`
```

Wanted: `MulAssign<f64>`, `DivAssign<f64>`, `MulAssign<i64>`, `DivAssign<i64>`.

Version: 4.3.1

### LINKED ISSUE #507: Python: duration / float raises TypeError
The binding defines `__div__`, which is a Python 2 dunder. PyO3 wires the `/`
operator to `__truediv__` only, so `__div__` is exposed as an ordinary method and
the operator never works.

```python
from hifitime import Duration, Unit

d = Unit.Hour * 1.0
d / 2.0            # TypeError: unsupported operand type(s) for /
d.__div__(2.0)     # works, but nobody writes this
```

`src/duration/python.rs` should define `__truediv__`.

Version: 4.3.1

greglucas:
## Exact `Duration` scaling by `f64`

Rewrites `Mul<f64>` and `Div<f64>` for `Duration` to be exact and total, plus the
surrounding gaps that turned up alongside them.

### The main bug

`Div<f64>` truncated the divisor to whole nanoseconds before dividing:

| expression | before | after |
|---|---|---|
| `1.hours() / 2.5` | 30 min | 24 min |
| `1.hours() / 1.5` | 1 h | 40 min |
| `1.hours() / 0.5` | panic | 2 h |
| `1.hours() / 0i64` | panic | `Duration::MAX` |

### How it works now

Both operators decompose `q` into its exact `mantissa * 2^exponent` form and apply
it in wide integer arithmetic, so the only rounding is the final truncation toward
zero to whole nanoseconds. Results saturate at `Duration::MAX` / `MIN`.

Neither panics:

| `q` | `d * q` | `d / q` |
|---|---|---|
| `±∞` | saturates by sign | `Duration::ZERO` |
| `±0.0` | `Duration::ZERO` | saturates by sign |
| `NaN` | `Duration::ZERO` | `Duration::ZERO` |

The infinity and zero rows are the mathematical limits and mirror each other. NaN
gives `ZERO` rather than a bound chosen by its sign bit; this matches the existing
`Freq * f64`.

### Commits

1. **Fix Duration scaling by f64** — the rewrite, plus `MulAssign`/`DivAssign`,
   a simplified `Mul<i64>`, and `debug_assert`ed preconditions on the wide-arithmetic
   helpers.
2. **Scale by a power of two with a shift** — `2.0`, `0.5`, `0.25` etc. have a
   mantissa of 1, so the scaling is one shift. `Duration / f64 (power of two)` goes
   38.7 ns → 12.5 ns; general paths within ~2%.
3. **`try_mul_f64` / `try_div_f64` + non-finite policy** — a `Result` API for the
   cases the operators substitute a sentinel, and `# Panics` sections on every
   panicking `f64` entry point.
4. **python: wire `/` to `__truediv__`**.

### Non-finite policy

The `f64` *constructors* still panic — that is deliberate, not an oversight: it's
covered by eight `#[should_panic]` tests in `tests/epoch.rs` and by 41
`#[kani::requires(...is_finite())]` preconditions, and it matches
`std::time::Duration::from_secs_f64`. Saturating there would silently turn
`Epoch::from_tai_seconds(f64::NAN)` into 1900-01-01, a plausible-looking date.

The split is now documented rather than changed, and `try_mul_f64` / `try_div_f64`
give the recoverable path for scaling. Fallible `Epoch` constructors are left for a
follow-up.

### Breaking changes

- `d / q` for non-integral `f64` `q` returns a different (correct) value.
- `d / 0.0` and `d / 0i64` saturate instead of panicking.
- `d * f64::NAN` returns `Duration::ZERO` instead of `MAX`/`MIN`.
- Python: `duration.__div__(...)` is gone; use `/`.

---

Closes #502, closes #503, closes #504, closes #505, closes #506, closes #507


ChristopherRabotin:
Thanks Greg! I'll do a detailed review this weekend!

ChristopherRabotin:
I haven't forgotten about this PR, just busy on other work at the moment, sorry.

